PR #32, opened 2015-06-04
conservative spell check run

SamPenrose:


hdashnow:
Much improved, thanks @SamPenrose!

Only thing is inconsistency between Australian and US English. I see color and vectorisation. I'm sure there are plenty of examples of this.

Presumably we should move everything to US English given the origins of SWC?

naupaka:
Thanks!

Issue #25, opened 2015-06-04
Pull in lesson-example CONTRIBUTING.md and LAYOUT.md


sritchie73:
This information has been updated since the repository has started. Key information includes R markdown usage. Pull from https://github.com/swcarpentry/lesson-example

aammd:
Agree that we need this.

aammd:
@jdblischak  mentioned over in #17  that 

> CONTRIBUTING.md is part of the core lesson-template repo. New changes to the lesson template are merged from this repo into downstream lesson repos like this one. Thus making changes to this file creates the risk of merge conflicts. 

I read this as meaning we need to treat the CONTRIBUTING.md over in lesson-template as the master copy, and use git to bring recent versions of those files over here, into the gapminder lesson. AM I understanding that correctly?

PR #36, opened 2015-06-04
adding changes to CONTRIBUTING.md from the lesson-template repo

aammd:
this addresses #25


jdblischak:
I'd recommend the following strategy instead:

```
git remote add lesson-template git@github.com:swcarpentry/lesson-template.git
git pull lesson-template gh-pages
```
This will pull not only the latest changes to CONTRIBUTING.md, but also all the other template files.

gvwilson:
+1 - this is what I do with other lessons.
Thx,
G

aammd:
Oh, I thought that a file by file process would be better ( and also thought that we would use lesson-example not lesson-template). I'll change that now.

aammd:
for the record, is it not better for one to push to a branch, then ask another person to look at it?

naupaka:
Thanks all!

naupaka:
Closes #25

jdblischak:
> for the record, is it not better for one to push to a branch, then ask another person to look at it?

Definitely. I think you took the right approach with this PR. When you are merging a lot of changes, it's often a good idea to create a feature branch and send a PR to get at least a second set of eyeballs. The only time I don't bother to send a PR is when I am doing something very routine like rendering an html file after some small changes have been made to an Rmd file.

hdashnow:
CONTRIBUTING.md currently says "For our lessons, you should branch from and submit pull requests against the gh-pages branch."

I think this works okay for smaller changes. Where it causes problems is when you want to do a large change, rearrange the order of lessons etc. So perhaps change the comment to something like:

"For most changes to our lessons, you should branch from and submit pull requests against the gh-pages branch.
For larger, or structural changes it's sometimes better to submit pull requests against a new branch for that feature. 
Don't worry, the topic maintainers will tell you when a change is big or contentious enough to need it's own branch. So just submit a pull request to get the conversation started."

And we should switch to this feature branch method after the sprint (I don't think we should mess with the momentum now).

hdashnow:
I should also say, we should make sure that newcomers to Git can easily make pull requests via GitHub so they don't have to use the command-line. The easiest way to do this is against default branch, in this case gh-pages.

aammd:
@hdashnow I'm not sure I understand. My git skills are OK but not expert at all, so I might be misunderstanding. what do you mean by 'pull request against' a branch? 
I ask because my understanding was that all contributors would submit pull requests *from* their own branches (`YourGithubName:informative-branch-name`) *to* this one (`swcarpentry:gh-pages`) Is that what you mean by "against"?
The only alternative, or so I thought, exists for maintainers. Because they have write access to this repository, they can push changes directly. I thought this is what @jdblischak was referring to?
Perhaps you could submit a PR for the change you suggest to CONTRIBUTING.md and we could discuss this there? :stuck_out_tongue_closed_eyes:

hdashnow:
Yes, the's the PR scenario that I'm talking about. I'm just saying you can do it would using the command-line.

If you don't know how to use the command-line git, you can do this:

* Go to https://github.com/swcarpentry/r-novice-gapminder
* Find the file you want to edit
* Click the edit button
* It will fork the repo, create a branch called "patch-1", and drop you in an editing window.
* Once done editing, it offers to let you make a pull request

This is super easy even for people who aren't familiar with git branching-merging workflows.

Is that the question you want answered? Or am I just going off on a tangent...?

aammd:
@hdashnow I think I understand your point of view better now! thanks :smile: 
We're going to be keeping the CONTRIBUTING and LAYOUT files here consistent with the [lesson-template](https://github.com/swcarpentry/lesson-template) repo (as Greg et al describe above). I think your suggestion could be a PR or issue on that repo.

jdblischak:
I apologize for creating any confusion during your sprint. I'll clarify what I was trying to say.

@aammd asked about using feature branches for PRs. Since he is a collaborator, and likely has to manage lots of changes to this repo, I agreed with his use of a feature branch when submitting a large change. This way, if it doesn't get merged immediately, he can continue to work on the main gh-pages branch. However, our contributing guidelines do not mention the use of feature branches. This is because they are difficult for novices, and unnecessary since they are likely only submitting one small change at any given time.

> For larger, or structural changes it's sometimes better to submit pull requests against a new branch for that feature.
Don't worry, the topic maintainers will tell you when a change is big or contentious enough to need it's own branch. So just submit a pull request to get the conversation started.

This was not what I was suggesting. PRs should be submitted against the gh-pages branch. The feature branch would only exist in the contributor's fork.

I think there are advantages and disadvantages to having contributors use the GitHub interface, but I think that discussion should take place on a different thread. I'd recommend either the discuss list or submitting an issue to the lesson-template repo.

aammd:
Thank you very much for the clarification, John. I am indeed a maintainer/contributor on this repo. The workflow you describe is how we are going to proceed for the near future. 
@hdashnow I'd encourage you to add your thoughts to the places that John suggested :smile:

PR #37, opened 2015-06-04
Alternate start

SamPenrose:
@mikabr and I (and maybe @hdashnow ) wanted to explore a more data- and demo-centric opening lesson. I grabbed a chunk of lesson 8; Mika made original contributions using dplyr and tidyr. The idea would be to make the existing 01 a new 02 (and so on). Questions:

- How does it feel to begin with a motivating example?
- Should we teach dplyr and/or tidyr, or should we replace those examples with something from packages already in the lesson?
- If we like this approach to lesson 1, should we move any of the later lessons towards this approach of less exposition, more payoff?

Finally, @aammd pointed out that we have an empty motivation.md slideshow which might serve as an alternate home for a gee-whiz demo.

### DIFF
--- a/xx-why-R.Rmd
+++ b/xx-why-R.Rmd
@@ -15,21 +15,22 @@ gapminder <- read.csv("data/gapminder-FiveYearData.csv", header=TRUE, sep=',')
 >
 > * To get a taste of R's powerful visualisation capabilities
 > * To get a taste of R's powerful statistical analysis capabilities
-> * To show how visualisation and analysis support each other
+> * To show how interweaving those capabilities pays off
 >
 
 ### Introduction to R
 
 Welcome to the R portion of the Software Carpentry workshop. We're going to show you how R and RStudio can help you understand large data sets. We'll also guide your first steps towards using them effectively for your own work.
 
-> #### Installation
-> (d/l RStudio)
-> (d/l gapminder -- can we include in this repo?}
-> install.packages('ggplot2')
-> (what else?)
+> ### Installation
+> * Download RStudio from [http://www.rstudio.com/products/rstudio/download/](http://www.rstudio.com/products/rstudio/download/)
+> * (Download gapminder data -- can we include in this repo?}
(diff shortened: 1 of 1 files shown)


sritchie73:
@SamPenrose I'm definitely in favour of starting out with a more data and demo-centric example. I think it's good to give students a quick win, show them how easy it is to do something in R when the stars align (all the data is clean, tidy, and none of R's esoteric behaviour comes up) before getting into the nitty-gritty of dealing with R in the real world.

hdashnow:
I like this idea!

You'll just need to move the section about how to download the gap-minder data up to the start of this lesson. 

I don't know if it's worth talking briefly about the format of the data right before you load it? If you did that, you could probably completely remove the loading data section from later lessons. Not sure if it would ruin the flow at this point though...

For the lesson timing, this will actually catch most of the technical difficulties. So leave plenty of time to stop and help students. A lot of red stickies go up the first time students use non-base packages in my experience!

aammd:
On the other hand, if this were the motivation slideshow, there would be no expectation that students follow along (if I'm understanding it correctly). Its more of a show-and-tell , "by tomorrow you can do this" thing.

hdashnow:
That's a good point. We generally run everything with students typing along. I suggest getting them to follow along with the easier parts of it (e.g. just loading the data and doing the plot), then we can pause to deal with the technical difficulties (I actually think it's good to catch this right at the start), and then ask students to just watch for the remainder of the intro.

aammd:
@hdashnow Oh definitely. I think that is exactly how SWC workshops ought to be done. the instructor codes on the screen and student's follow along. 

These motivational slides ([See the LAYOUT.md file here](https://github.com/swcarpentry/r-novice-gapminder/blob/gh-pages/LAYOUT.md#motivational-slides)) are only supposed to be short ("3 minutes or less"). Nobody follows along, they just watch. 

Just so that we're all on the same page. :smile_cat:

hdashnow:
Let's do it and trial it at the next workshop!

PR #44, opened 2015-06-05
Archival PR: materials used in 2015-04-27 workshop at Washington State University (do not merge)

karawoo:
These are the materials we used for the R lessons at Washington State University on 2015-04-27. We didn't get through the control flow section, and I think we ended up cutting the ggplot2 lesson a little short.


karawoo:
@aammd could you add the `workshop-archive` label? I don't think I have the permissions.

aammd:
Thank you @karawoo ! This is great :+1:

Issue #40, opened 2015-06-05
Add motivational slides


aammd:
We ought to have a short [Motivational slideshow](https://github.com/swcarpentry/r-novice-gapminder/blob/gh-pages/LAYOUT.md#motivational-slides) that we can use at the start of a lesson. from the LAYOUT.md : 

> Every lesson must include a short slide deck in motivation.md suitable for a short presentation (3 minutes or less) that the instructor can use to explain to learners how knowing the subject will help them. 

I was thinking that #37 would make really great material for such a thing. The original authors could work on that, or if people agree we can merge that in and someone else can make it into slides 
cc @hdashnow @SamPenrose

SamPenrose:
I am working on this today.

naupaka:
It looks like #37 and #45 have been merged and so can we close this issue?

aammd:
Indeed! closing.

PR #45, opened 2015-06-05
PR#37 converted to slides per issue #40

SamPenrose:
The change in format from read-along lesson to slideware had a larger and more negative impact than I expected, probably because slides are so painfully low-resolution. @mikabr, I think your introduction of dyplr and tidyr suffered especially. Contributors, let me know what you think.


aammd:
Merging this PR because I think we need a motivation.html slide deck, and this is an excellent start! However, I deeply love `dplyr` and `tidyr`, and so I don't want this material to suffer at all. the original dplyr & tidyr part is around somewhere, right?

Issue #26, opened 2015-06-04
Add general description and layout of functions


hdashnow:
The start of 07-functions lesson needs some text about what a function is, and it's general form (like the if/else lesson).

SamPenrose:
I am going to look at the SWC Python and Shell repos to see if they have something to reuse.

naupaka:
Fixed by #46

PR #46, opened 2015-06-05
callout defining function

SamPenrose:
To address Issue #26

### DIFF
--- a/07-functions.Rmd
+++ b/07-functions.Rmd
@@ -29,6 +29,18 @@ from a different source in the future.
 In this lesson, we'll learn how to write a function so that we can repeat
 several operations with a single command.
 
+> ## What is a function? {.callout}
+>
+> Functions gather a sequence of operations into a whole, preserving it for ongoing use. Functions provide:
+>
+> * a name we can remember and invoke it by
+> * relief from the need to remember the individual operations
+> * a defined set of inputs and expected outputs
+> * rich connections to the larger programming environment
+>
+> As the basic building block of most programming languages, user-defined functions constitute "programming" as much as any single abstraction can. If you have written a function, you are a computer programmer.
+>
+
 
 ### Defining a function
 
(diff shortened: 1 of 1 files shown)


SamPenrose:
p.s., I also wrote this (per the issue), but I don't actually think its an improvement over just getting into the example:

Functions take the form:

*function_name* <- **function(** *parameters* **)** *function_body*

where:

- *function_name* what the programmer has decided to call her function
- *parameters* is an optional list of named values the function accepts as input
- function_body is a sequence of expressions

You invoke a function by entering its name followed by the call operator, a pair of parentheses: "()". If the function takes parameters, they go inside the parentheses. R will then execute the function's body, expression by expression. If you assign the result of the function to a variable:

```
x = some_function('some', 'parameters')
```

then, when the function finishes executing, whatever it *returns* will be stored in that variable. By default, R functions return the last evaluated expression. Many functions use the *return* keyword to send a specific object back to the caller.

aammd:
I really like this Callout! I made a few line notes. 

I have pretty mixed feelings about this "imitation syntax" approach that I've seen sometimes:

> function_name <- function( parameters ) function_body

Because it doesn't really resemble actual syntax (ie no `{`), I'm not sure it doesn't generate more confusion than it solves.

sritchie73:
I think its a really good idea to have this callout: I'd even suggest putting it in the very first topic after the first instance where an instructor calls a function. 

I'm also not a fan of the "imitation syntax". Novices seem to struggle with pseudocode, its too abstract for them to understand, especially since they've been hammered with lots of new concepts.

hdashnow:
I think showing the general form of a function is pretty important. But I agree pseudocode can be confusing for beginners. How about something like this? It's still very general, but at least the code would run.

Functions take the form:

```python
function_name <- function( input ) {
    result <- input + 1 # The body of the function. 
    return( result ) # Optional return statement
}
```

sritchie73:
I like it!

aammd:
I really prefer this. Now we have a function that is instructive for beginners, while not being distressing to people who already write functions. I will resolve whatever conflicts (!) we have here and merge this.

PR #50, opened 2015-07-15
Update 05-data-structures-part2.md

jeremycg:
The output of the class(df$id) at line 79 should be "factor" - we are deliberately introducing this error to solve it below.


jeremycg:
apologies, pulled to .md instead of .rmd

PR #54, opened 2015-07-23
draft lesson for dplyr and tidyr

remi-daigle:
Hey @aammd,

Here is a draft version of my dplyr lesson, let me know what you think. I'm working on tidyr at the moment. We may not want to merge this fork yet? But somehow when I started this was the easiest way to get this to your attention!


sritchie73:
+1 this looks very nice!

remi-daigle:
OK, I just finished putting together another lesson for tidyr, I 'think' I added it to this PR

13-dplyr.Rmd
```diff
@@ -0,0 +1,191 @@
...
+```{r, include=FALSE}
+source("tools/chunk-options.R")
+opts_chunk$set(fig.path = "fig/12-plyr-")
+# Silently load in the data so the rest of the lesson works
+gapminder <- read.csv("data/gapminder-FiveYearData.csv", header=TRUE)
+```
+
+> ## Learning objectives {.objectives}
+>
+> * To be able to use 6 main dataframe manipulation 'verbs' with pipes in `dplyr`
```
naupaka: Add "the" before 6

13-dplyr.Rmd
```diff
@@ -0,0 +1,191 @@
...
+>
+
+Manipulation of dataframes means many things to many researchers, we often select certain observations (rows) or variables (columns), we often group the data by a certain variable(s), or we even calculate summary statitistic. We can do these operations using the normal base R operations
```
naupaka: "statistic" -> "statistics" and add a colon after "operations"

13-dplyr.Rmd
```diff
@@ -0,0 +1,191 @@
...
+mean(gapminder[gapminder$continent == "Asia", "gdpPercap"])
+```
+
+But this isn't very *nice* because there is a fair bit of repetition. Repeating yourself will cost you time, both now and later, and potentially introduce some nasty bugs.
+
+#### The `dplyr` package
+
+Luckily, the [`dplyr`](https://cran.r-project.org/web/packages/dplyr/dplyr.pdf) package provides a number of very useful functions for manipulating dataframes in a way that will reduce the above repetition, reduce the probability of making errors, and probably even save you some typing. As an added bonus, you might even find the `dplyr` grammar easier to read.
+
+Here we're going to cover 6  of the most commonly used functions as well as using pipes (`%>%`) to combine them.
```
naupaka: Extra space after 6

13-dplyr.Rmd
```diff
@@ -0,0 +1,191 @@
...
+
+#### Using select()
+
+If, for example, we wanted to move forward with only a few of the variables in our dataframe we could use the `select` function. This will keep only the variables you select.
+
+```{r}
+year_country_gdp <- select(gapminder,year,country,gdpPercap)
+```
+
+> draw this on the board? I'd like to include a diagram here too
```
naupaka: This seems like an aside to the instructor? Should tweak to be an aside (or remove?) before lesson is merged?

13-dplyr.Rmd
```diff
@@ -0,0 +1,191 @@
...
+
+If we open up `year_country_gdp` we'll see that it only contains the year, country and gdpPercap. Above we used 'normal' grammar, but `dplyr` strengths lie in combining several functions using pipes. Since the pipes grammar is unlike anything we've seen in R before, let's repeat what we've done above using pipes.
```
naupaka: `dplyr` strengths lie -> the strengths of `dplyr` lie

naupaka: Could also mention the similarity of the R pipes concept to the pipes concept in the shell lessons (which they may or may not have seen before this lesson, although probably they will have, since this is a more advanced topic?)

13-dplyr.Rmd
```diff
@@ -0,0 +1,191 @@
...
+
+```{r}
+year_country_gdp <- gapminder %>% select(year,country,gdpPercap)
+```
+
+> draw this on the board? I'd like to include a diagram here too
```
naupaka: Same comment as above - make a formal instructor box for this aside or remove it?

13-dplyr.Rmd
```diff
@@ -0,0 +1,191 @@
...
+
+To help you understand why we wrote that in that way, let's walk through it step by step. First we summon the gapminder dataframe and pass it on, using the pipe symbol `%>%`, to the next step which is the select function. In this case we don't specify which data object we use in the `select` function since in gets that from the previous pipe.
```
naupaka: select -> `select`

13-dplyr.Rmd
```diff
@@ -0,0 +1,191 @@
...
+
+```{r}
+year_country_gdp_euro <- gapminder %>%
+    filter(continent=="Europe") %>%
+    select(year,country,gdpPercap)
+```
+
+> #### Challenge 1 {.challenge}
+>
+> Can you write a single command (which can span multiple lines and includes pipes) that will produce a dataframe that has the African values for `lifeExp`, `country` and `year`, but not for other Continents.
```
naupaka: Can you write -> Write

13-dplyr.Rmd
```diff
@@ -0,0 +1,191 @@
...
+```
+
+> #### Challenge 1 {.challenge}
+>
+> Can you write a single command (which can span multiple lines and includes pipes) that will produce a dataframe that has the African values for `lifeExp`, `country` and `year`, but not for other Continents. 
+>How many rows does your dataframe have and why?
+>
+
+
+As with last time, first we pass the gapminder dataframe to the filter function, then we pass the filtered version of the gapminder dataframe to the select function. **Note:** The order of operations is very important in this case. If we used 'select' first, filter would not be able to find the variable continent since we would have removed it in the previous step.
```
naupaka: filter -> `filter` and select -> `select`

naupaka: Also, many times the best practice for R functions is to include the parentheses after the function name, so, e.g., `select` is written as `select()`

13-dplyr.Rmd
```diff
@@ -0,0 +1,191 @@
...
+
+#### Using group_by()
+
+Now, I told you all that I would be reducing the error prone repetitiveness of what we can do with base R, but up to now we haven't done that since we would have to repeat the above for each continent. Instead of `filter`, which will only pass observations that meet your criteria (in the above: `continent=="Europe"`), we can use `group_by`, which will essentially use every unique criteria that you could have used in filter.
```
naupaka: Should be written not as a transcript to be read, but as a description of the information to be taught.  I think also perhaps these few sentences could be reworded for clarity. I think this is also true for the explanation on line 102. The `group_by()` function is a bit of a tricky concept to explain well and so could potentially use a bit more space to explain?

13-dplyr.Rmd
```diff
@@ -0,0 +1,191 @@
...
+
+```{r}
+year_country_gdp_bycontinents <- gapminder %>%
+    group_by(continent) %>%
+    select(year,country,gdpPercap)
+```
+You will have noticed that `year_country_gdp_bycontinents` is the same as `year_country_gdp` where we had just used `select` and no `group_by`. That's because we essentially passed a dataframe for each continent to the `select` but then we just put those 5 dataframe into `year_country_gdp_bycontinents`. So in other words, we grouped them, them ungrouped them without doing anything special except `select` which we could have done without grouping.
```
naupaka: no -> not

naupaka: to the `select` -> to `select()`

naupaka: I think this is a bit of a confusing example to show off `group_by()`. Maybe there's a more straightforward way to show this?

13-dplyr.Rmd
```diff
@@ -0,0 +1,191 @@
...
+
+#### Using summarize()
+
+The above was a bit on the uneventful side because `group_by` works much better in conjunction with `summarize`, which is that 'something special' I alluded to above.
```
naupaka: Why not get to this in conjunction with `group_by()` and cut straight to the point? No real reason to use `group_by()` without something like `summarize()` afterwards.

13-dplyr.Rmd
```diff
@@ -0,0 +1,191 @@
...
+
+```{r}
+gdp_bycontinents <- gapminder %>%
+    group_by(continent) %>%
+    summarize(mean_gdpPercap=mean(gdpPercap))
+```
+
+
+Now that's getting exciting! That allowed us to calculate the mean gdpPercap for each continent, but it gets even better! Let's set we wanted to also group by year
```
naupaka: also group by year -> also group by year...

13-dplyr.Rmd
```diff
@@ -0,0 +1,191 @@
...
+
+> #### Challenge 2 {.challenge}
+>
+> Calculate the average life expectancy per country. Which had the
+> longest and shortest?
```
naupaka: Change to: "which had the longest life expectancy and which had the shortest life expectancy?"

13-dplyr.Rmd
```diff
@@ -0,0 +1,191 @@
...
+              sd_pop=sd(pop))
+```
+
+
+
+
+
+
+> #### Advanced Challenge {.challenge}
+> Calculate the average life expectancy in 2002 of 2 randomly selected countries for each continent. Then arrange the continent names in reverse order
```
naupaka: add period at end of sentence (after order)

13-dplyr.Rmd
```diff
@@ -0,0 +1,191 @@
...
+> **Hint:** Use the `dplyr` functions `arrange()` and `sample_n()`, they have similar syntax to other dplyr functions
```
naupaka: Add a period here too.

14-tidyr.Rmd
```diff
@@ -0,0 +1,154 @@
...
+```{r, include=FALSE}
+source("tools/chunk-options.R")
+opts_chunk$set(fig.path = "fig/12-plyr-")
+# Silently load in the data so the rest of the lesson works
+gapminder <- read.csv("data/gapminder-FiveYearData.csv", header=TRUE)
+```
+
+> ## Learning objectives {.objectives}
+>
+> * To be able to convert dataframes between 'long' and 'wide' format with `tidyr`
```
naupaka: Do they also need to know what these mean? I would include that in the learning objectives as well. Many people may not be familiar with these concepts yet (unless they were specifically covered earlier).

14-tidyr.Rmd
```diff
@@ -0,0 +1,154 @@
...
+Researchers often want to manipulate their data from the 'wide' to the 'long' format, or vice-versa. The 'long' format is where:
+
+ - each column is a variable
+ - each row is an observation
+
+In the 'long' format, you usually have 1 column for the observed variable and the other columns are ID variables. 
+
+For the 'wide' format each row is often a site/subject/patient and you have multiple observation variables. These can be either repeated observations over time, or observation of multiple variables (or a mix of both). You may find data input may be simpler or some other applications may prefer the 'wide' format. However, many of `R`'s functions have been designed assuming you have 'long' format data. This tutorial will help you efficiently transform your data regardless of original format.
+ 
+> draw this on the board? I'd like to include a diagram here too
```
naupaka: Same comment as earlier about this aside note to the instructor. Should be formalized or removed before merging.

14-tidyr.Rmd
```diff
@@ -0,0 +1,154 @@
...
+>
+
+Researchers often want to manipulate their data from the 'wide' to the 'long' format, or vice-versa. The 'long' format is where:
+
+ - each column is a variable
+ - each row is an observation
+
+In the 'long' format, you usually have 1 column for the observed variable and the other columns are ID variables. 
+
+For the 'wide' format each row is often a site/subject/patient and you have multiple observation variables. These can be either repeated observations over time, or observation of multiple variables (or a mix of both). You may find data input may be simpler or some other applications may prefer the 'wide' format. However, many of `R`'s functions have been designed assuming you have 'long' format data. This tutorial will help you efficiently transform your data regardless of original format.
```
naupaka: I think a little more time could be spent here explaining the differences between wide and long, and why they are useful (e.g. discussing how this format is more machine readable, how it shares many similarities with database formatting of fields and values within tables, etc).

14-tidyr.Rmd
```diff
@@ -0,0 +1,154 @@
...
+library("tidyr")
+library("dplyr")
+```
+
+First, lets look at the structure of our gapminder dataframe:
+```{r}
+str(gapminder)
+```
+
+It is somewhere in between the purely 'long' and 'wide' format. We have 3 "ID variables" (`continent`, `country`, `year`) and 3 "Observation variables" (`pop`,`lifeExp`,`gdpPercap`). I usually prefer my data in this format in most cases despite not having ALL observations in 1 column. Since all 3 observation variables have different units, there are not too many operations that would need us to stretch out this dataframe any longer (i.e. 4 ID variables and 1 Observation variable). Many or R's vector based operations like oav(), lm(), or glm() will be easiest in this intermediate format (i.e. long format for each dependent variable, but not the longest format).
```
naupaka: Not sure many novice learners will have encountered these modeling functions. Maybe either worth spending a little more time explaining the type of function that prefer this intermediate formats or else remove that to keep focused?

14-tidyr.Rmd
```diff
@@ -0,0 +1,154 @@
...
+
+However, converting to the pure 'long' format is the easiest way to convert this data to the pure 'wide' format. Manually entering the data into a spreadsheet is more intuitive in the wide format and some plotting functions actually work better in the wide format.
+
+Before we really start, let's re-organize the ID and Observations variables so that they are on order (this will make more sense later). We're swapping `pop` and `continent`
```
naupaka: Why are we swapping the order?

14-tidyr.Rmd
```diff
@@ -0,0 +1,154 @@
...
+```{r}
+gap_long <- gapminder %>% gather(obs_type,obs_values,-continent,-country,-year)
+str(gap_long)
+```
+
+That may seem trivial with this particular dataframe, but sometimes you have 1 ID variable and 40 Observation variables. The flexibility is a huge time saver!
+
+> #### Challenge 1 {.challenge}
+>
+> Using `gap_long`, can you calculate the mean life expectancy, population, and gdpPercap for each continent?
```
naupaka: "can you calculate" -> "calculate"
Remove question mark at the end.

14-tidyr.Rmd
```diff
@@ -0,0 +1,154 @@
...
+>**Hint:** use the `group_by()` and `summarize()` functions we learned in the dplyr lesson
```
naupaka: dplyr -> `dplyr` 
Important to be consistent in the use of monospaced code font

14-tidyr.Rmd
```diff
@@ -0,0 +1,154 @@
...
+Now just to double-check our work, lets use the opposite of `gather()` to spread our observation variables back out with the aptly named `spread()`
+
+```{r}
+gap_normal <- gap_long %>% spread(obs_type,obs_values)
+identical(gap_normal,gapminder)
+```
+
+That's great! We've gone from the longest format back to the intermediate and we didn't introduce any errors in our code.
+
+Now lets convert the long all the way to the wide. In the wide format, we will keep country and continent as ID variables and spread the observations across the 3 metrics (`pop`,`lifeExp`,`gdpPercap`) and time (`year`). First we need to create appropriate labels for all our new variables (time*metric combinations) and we also need to unify our ID variables to simplify the process of defining `gap_wide`
```
naupaka: lets -> let's

naupaka: I think this explanation is still a bit complicated. Maybe simplify a bit further?

14-tidyr.Rmd
```diff
@@ -0,0 +1,154 @@
...
+gap_wide <- gap_long %>% 
+    unite(ID_var,continent,country,sep="_") %>%
+    unite(var_names,obs_type,year,sep="_") %>%
+    spread(var_names,obs_values)
+str(gap_wide)
+```
+
+> #### Challenge 2 {.challenge}
+>
+> Can you take this 1 step further and create a `gap_ludicrously_wide` format data by spreading over countries, year and the 3 metrics?
```
naupaka: "Can you take this" -> "Take this"

14-tidyr.Rmd
```diff
@@ -0,0 +1,154 @@
...
+>**Hint** this new dataframe should only have 5 rows.
+>
+
+Now we have a great 'wide' format dataframe, but the `ID_var` could be more usable, let's separate them into 2 variables with `separate()`
```
naupaka: Pronoun disagreement. Change "them" to "it" or add "column" after `ID_var` and then change "them" to "it".

naupaka:
Thanks for the PR, @remi-daigle! These will be great to have included as lessons after a few minor tweaks. I've gone through and added line comments on the files to hopefully help clarify a few things.

remi-daigle:
Hey @naupaka Thank you for the great feedback! I'll fix this up ASAP, but until then, does anyone know a good way of generating nice dataframe diagrams. Ideally I'd love to do something more reproducible than drawing something in photoshop or something, making a .jpg and incorporating that into the lesson

naupaka:
@remi-daigle maybe try markdown tables? https://help.github.com/articles/github-flavored-markdown/ There are a few packages out there (`kable`, `xtable` etc) that allow you to format your dataframes as md or html or tex tables.

remi-daigle:
@naupaka I _could_ use those, but I was looking to plot more of a 'toy' version of the dataframes, I would prefer if the didn't actually include all the data. I'd like to do things like highlight rows or columns and have an arrow going to other table(s). This would be more to illustrate what's happening with the row/columns rather than displaying what's actually in them. It looks like the diagram package would give me the arrows, but not sure how to make the toy tables to go with that. Does that make sense?

naupaka:
It does. I'm afraid I don't know of a good solution off the top of my head then. But I can see why you'd want one... I think this was the idea of the `rblocks` package? [link here](http://ramnathv.github.io/posts/rblocks/).

remi-daigle:
`rblocks` looks pretty good, it sounds better than what I'm attempting! I wish it was further along in the development phase. I think I'm getting there with the DiagrammeR package and some html (learning html in the process!). It looks like it will do what I need it to. A quick sample:
image

aammd:
That's cool! You might also try drawing them using `gridExtra::tableGrob()`, [as shown here](http://stackoverflow.com/questions/12318120/adding-table-within-the-plotting-region-of-a-ggplot-in-r)

remi-daigle:
OK, I think I have addressed all of @naupaka 's comments, although the tidyr lesson still has some issues because A) I'm having some super basic issue with reading a writing csv's leading to non-identical dataframes, not sure what I'm doing wrong! (see this [commit](https://github.com/remi-daigle/r-novice-gapminder/commit/4e2d790bf504fde6a6fbf09fdcbe59ddaf58555b) and B) I restructured the lesson quite a bit some some new issues may have popped up

13-dplyr.Rmd
```diff
@@ -0,0 +1,578 @@
...
+
+If, for example, we wanted to move forward with only a few of the variables in our dataframe we could use the `select()` function. This will keep only the variables you select.
+
+```{r}
+year_country_gdp <- select(gapminder,year,country,gdpPercap)
+```
+
+```{r,echo=FALSE}
+library(DiagrammeR)
+grViz('digraph html {
```
aammd: @remi-daigle this is awesome! I love that you found a way to create these illustrations. Do you think it would be easier to read and maintain these .Rmd files if these long `DiagrammeR` commands were kept in the `tools/` directory? @naupaka what do you think?

remi-daigle:
Hey @aammd I'm not opposed at all to moving them to the tools/ directory. Those commands are definitely huge (though deceptively simple once I 'got' it) and they negatively affect the readability of the .Rmd files.

My first instinct was to keep everything self contained at first, but if there's a better way, I'm all for it! Perhaps an Rscript for each of the figures? That way we could just source a script for each figure? Either way, let me know!

13-dplyr.Rmd
naupaka: I think so... otherwise it is a lot of material to scan through... `source(figure1.R)` etc would be a lot more concise. But I am not sure what best practice should be. Maybe @gvwilson or @jdblischak could weigh in?

naupaka:
What does the diff look like between the two `csv` versions? Could it be a unicode/encoding thing for an apostrophe or something like that?

remi-daigle:
The diff won't make sense since the `csv`'s are formatted differently (like number of columns and rows differ). What I did was load the original `gapminder` dataset, I manipulated the original data into the 'wide' format with `dplyr` and assigned it to `gap_wide`. Now the whole point of the lesson is to go from wide, to long, then to intermediate (which is the original `gapminder` format). So in the end, their dataframe *should* be identical to the original.

If I use the `gap_wide` that I generated from the original `gapminder` data with `dplyr` (which I don't want the students to have to do) everything works fine and their dataframe ends up being identical to the original `gapminder`. Trouble is, if I try to save `gap_wide` as a `.csv` and then read it back in (as the students would do). Their dataframe at the end is not identical to the original (despite passing the `all.equal()` test)

13-dplyr.Rmd
```diff
@@ -0,0 +1,578 @@
...
+# Silently load in the data so the rest of the lesson works
+gapminder <- read.csv("data/gapminder-FiveYearData.csv", header=TRUE)
+```
+
+> ## Learning objectives {.objectives}
+>
+> * To be able to use the 6 main dataframe manipulation 'verbs' with pipes in `dplyr`
+>
+
+Manipulation of dataframes means many things to many researchers, we often select certain observations (rows) or variables (columns), we often group the data by a certain variable(s), or we even calculate summary statistic. We can do these operations using the normal base R operations:
```
naupaka: statistic -> statistics

13-dplyr.Rmd
```diff
@@ -0,0 +1,578 @@
...
+')
+```
+
+If we open up `year_country_gdp` we'll see that it only contains the year, country and gdpPercap. Above we used 'normal' grammar, but the strengths of `dplyr` lie in combining several functions using pipes. Since the pipes grammar is unlike anything we've seen in R before, let's repeat what we've done above using pipes.
+
+```{r}
+year_country_gdp <- gapminder %>% select(year,country,gdpPercap)
+```
+
+To help you understand why we wrote that in that way, let's walk through it step by step. First we summon the gapminder dataframe and pass it on, using the pipe symbol `%>%`, to the next step which is the `select()` function. In this case we don't specify which data object we use in the `select()` function since in gets that from the previous pipe. **Fun Fact**: There is a good chance you have encountered pipes before in the shell. In R, a pipe symbol is `%>%` while in the shell it is `|` but the concept is the same!
```
naupaka: add a comma after 'the next step'

13-dplyr.Rmd
```diff
@@ -0,0 +1,578 @@
...
+#### Using group_by() and summarize()
+
+Now, we were supposed to be reducing the error prone repetitiveness of what can be done with base R, but up to now we haven't done that since we would have to repeat the above for each continent. Instead of `filter()`, which will only pass observations that meet your criteria (in the above: `continent=="Europe"`), we can use `group_by()`, which will essentially use every unique criteria that you could have used in filter.
+
+```{r}
+str(gapminder)
+
+str(gapminder %>% group_by(continent))
+```
+You will notice that the structure of the dataframe where we used `group_by()` (`grouped_df`) is not the same as the original `gapminder` (`data.frame`). A `grouped_df` can be thought of as a `list` where each item in the `list`is a `data.frame` which contains only the rows that correspond to the a particular value `continent` (at least in the example above)
```
naupaka: Add a period at the end of the line

13-dplyr.Rmd
```diff
@@ -0,0 +1,578 @@
...
+```
+
+That allowed us to calculate the mean gdpPercap for each continent, but it gets even better.
+
+> #### Challenge 2 {.challenge}
+>
+> Calculate the average life expectancy per country. Which had the longest life expectancy and which had the shortest life expectancy?
+>
+
+The function `group_by()` allows us to group by multiple variables. Let's group by `year` and `continent`
```
naupaka: Add period

14-tidyr.Rmd
```diff
@@ -0,0 +1,597 @@
...
+str(gap_wide_betterID)
+
+identical(gap_wide,gap_wide_betterID)
+```
+
+There and back again!
+
+> #### Solution to Challenge 1 {.challenge}
+>
+> The original gapminder data.frame is in an intermediate format. It is not purely long since it had multiple observation variables ()
```
naupaka: missing variable names between parentheses?

naupaka:
@remi-daigle I guess I am still confused. So it is equal but it is not equal? If you write out the final (back to original - I think you are calling this intermediate) dataset that you generated fully within R (that is, to `gap_wide` and back again, without using `write.csv()`) to `csv` and then compare this to the version that you *did* write out to `csv` (as the students would use) before converting back int the intermediate (original) form, what does the diff look like? This should be a diff of original (intermediate) data to original intermediate data (but one done entirely within R and the other written out to `csv` and then back in in an intermediate step). Whew that is confusing! Do you get what I mean though?

remi-daigle:
Yeah, I think I get what you mean, but I think the problem is something to do with the actual writing of the file as you suggested. I simplified the problem to something easily reproducible:

    #load the original data
    gapminder <- read.csv("data/gapminder-FiveYearData.csv", header=TRUE, stringsAsFactors =   FALSE)
    library(dplyr)
    library(tidyr)
    #transform the data to the wide format
    gap_wide <- gapminder %>% gather(obs_type,obs_values,-continent,-country,-year) %>% 
        unite(ID_var,continent,country,sep="_") %>%
        unite(var_names,obs_type,year,sep="_") %>%
        spread(var_names,obs_values) %>% 
        separate(ID_var,into=c('continent','country'),sep='_')

    #write our the .csv so students can use it
    write.csv(gap_wide_betterID,"data/gapminder_wide.csv",row.names = FALSE)
    #load the "student" data
    gap_wide2 <- read.csv("data/gapminder_wide.csv", header=TRUE, stringsAsFactors = FALSE)

    all.equal(gap_wide,gap_wide2)  #this returns TRUE
    identical(gap_wide,gap_wide2)  #this returns FALSE, but WHY?!

for the "students", the data.frames at the end of the lesson have the same problem (all equal, but not identical) presumably because they start off different as in the example above. As you said, there's a good chance it has something to do with line endings (I'm on a Windows machine right now and that's all I have access to at the moment since I'm moving), but I'm not sure what to do about it!

naupaka:
Somewhere, somehow, a floating point is upset. I think it is fine to just not use `identical()` to check in this case -- it's meant for logical comparisons primarily, I believe. 

From `?identical`

------
A call to identical is the way to test exact equality in if and while statements, as well as in logical expressions that use && or ||. In all these applications you need to be assured of getting a single logical value.

Users often use the comparison operators, such as == or !=, in these situations. It looks natural, but it is not what these operators are designed to do in R. They return an object like the arguments. If you expected x and y to be of length 1, but it happened that one of them was not, you will not get a single FALSE. Similarly, if one of the arguments is NA, the result is also NA. In either case, the expression if(x == y).... won't work as expected.

The function all.equal is also sometimes used to test equality this way, but was intended for something different: it allows for small differences in numeric results.

naupaka:
Ok, I'm happy to merge as-is or else to have the diagrams pulled out into separate files that are incorporated with `source()`. The latter would clutter up the directory with files, so I'm leaning towards perhaps just having the digram code stay in the single file as it is now. Final thoughts @aammd, before I push the BigGreenButton?

remi-daigle:
Awesome, Thanks for all the help @naupaka ! I'm fine with using all.equal() since it demonstrates what we are trying to show, although it still bothers me that identical() does not work...

For the `source()` option (if we go that route), is there a way to get the plots to display in the `.html` document without first saving to a `.jpg` when using `source()`. I have been toying with the options in `source()` and in the R chunk, but haven't found the magical combination yet. Having a bunch of `.jpg` would clutter 'tool/` even more so I would definitely want to avoid that route. I'll wait for @aammd 's final thoughts either way.

13-dplyr.Rmd
jdblischak: Seems reasonable to me to have these in a separate location to improve readability. Since they're figures, I'd recommend putting them in `fig`.

Also, I am not familiar with this library, so this may be trivial to implement. But when I put the `grViz` call in a separate R file and loaded it with `source`, the html file did not display the figure.

remi-daigle: @jdblischak @naupaka @aammd I don't know how to make it work with the `source()` method and I don't think I will be able to fix it. I can't seem to get it to print directly or to export it to file and then load it. It has something to do with the `shiny` package since the figures I believe are created as html widgets? Anyway, it's beyond my expertise to find a solution, any suggestions?

naupaka: I think we should just leave them in to keep everything simple. In fact, I'll merge it now since I think this was the last issue we were discussing.

remi-daigle:
Awesome, I see the new .Rmd files are now in the main repo, but when/how to they get up on the website?

naupaka:
Hi @remi-daigle the generated pages are now up [here](http://swcarpentry.github.io/r-novice-gapminder/index.html). As you'll notice, unfortunately, the graphviz elements are not rendering properly via jekyll. They work for me fine when I run it locally in RStudio, but not when the site is generated with pandoc. Can you look into how to fix that problem? Otherwise I think it would be best to swap in png files for the figures just to make sure it works...

PR #58, opened 2015-08-24
Added `str` to list of data type interrogations

rmcd1024:
Seemed helpful to discuss `str()` at the outset.

### DIFF
--- a/04-data-structures-part1.Rmd
+++ b/04-data-structures-part1.Rmd
@@ -43,6 +43,7 @@ is.numeric() # is it numeric?
 is.integer() # is it an integer?
 is.complex() # is it complex number data?
 is.character() # is it character data?
+str()  # what is it?
 ```
 
 > #### Challenge 1: Data types {.challenge}


04-data-structures-part1.Rmd
```diff
@@ -43,6 +43,7 @@ is.numeric() # is it numeric?
 is.integer() # is it an integer?
 is.complex() # is it complex number data?
 is.character() # is it character data?
+str()  # what is it?
```
naupaka: There's also `class()` -- but I think of `str()` and `class()` as more general interrogation functions than just for atomic types and it might get a tad confusing. E.g.:
```r
> a <- 1
> str(a)
 num 1
> class(a)
[1] "numeric"
> typeof(a)
[1] "double"
```

Is this the best place to have them introduced in the lesson? I definitely agree they should be in there somewhere though, and this indeed might be the best place to do it.

rmcd1024: Two thoughts:

1. Both when i read this and taught this, this seemed like an appropriate
place to introduce str(x). One could disagree. class() is not used at all
in this lesson.

2. The tutorial does use str(x) later, but without much explanation. So
this provides a prior opportunity to explain, and explain that there are
more general interrogation methods.



On Sun, Aug 23, 2015 at 9:34 PM, Naupaka Zimmerman <notifications@github.com

naupaka:
Thanks @rmcd1024!

PR #59, opened 2015-08-24
Change to data frames/lists tutorial

rmcd1024:
Modifications to the data frames/lists tutorial
    
    Among the changes:
    1. Emphasize cbind rather than rbind, with the idea that column
    additions are likely to be more common
    2. Show how to add a factor level so that rbind example will work
    3. Challenge covering `names` vs `colnames`
    
    The discussion of the gapminder csv file remains, even though there is
    now a gapminder package. It would probably be good to change to using the
    package.

### DIFF
--- a/05-data-structures-part2.Rmd
+++ b/05-data-structures-part2.Rmd
@@ -87,7 +87,9 @@ To add a row we use `rbind`:
 df <- rbind(df, list("g", 11, 42, 0, "G"))
 ```
 
-This doesn't work as expected! What do the error messages tell us?
+Note that we add the row as a list, because we have multiple types
+across the columns. Nevertheless, this doesn't work as expected! What
+do the error messages tell us?
 
 It appears that R was trying to append "g" and "G" as factor
 levels. Why? Let's examine the first column.  We can access a column
--- a/05-data-structures-part2.Rmd
+++ b/05-data-structures-part2.Rmd
@@ -462,11 +462,6 @@ summary(reg)
 As you might expect, life expectancy has slowly been increasing over
 time, so we see a significant positive association!
 
-> #### Challenge 6 {.challenge}
->
-> Write this function: `lm(lifeExp ~ year, data=gapminder)` without
-> using the `data` argument. Verify that you get the same answer.
-
 
 ## Challenge Solutions
 
@@ -501,7 +496,3 @@ time, so we see a significant positive association!
 > colnames(m) <- letters[1:3] # works as you would expect
 > names(m) <- letters[1:3]  # destroys the matrix
 > ```
-
-> #### Solution to Challenge 6 {.challenge}
-> 
-> `lm(gapminder$lifeExp ~ gapminder$year)`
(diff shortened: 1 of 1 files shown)


05-data-structures-part2.Rmd
```diff
@@ -330,44 +423,54 @@ Lets run a basic linear regression on the gapminder dataset:
...
+summary(reg)
 ```
 
 As you might expect, life expectancy has slowly been increasing over
 time, so we see a significant positive association!
 
+> #### Challenge 6 {.challenge}
+>
+> Write this function: `lm(lifeExp ~ year, data=gapminder)` without
+> using the `data` argument. Verify that you get the same answer.
```
naupaka: This is a **bad idea** because it will mess up later attempts to use the model for prediction. I would take this specific challenge out.

naupaka:
Other than the one in-line comment I made, I think this is a nice set of improvements and I am happy to merge. Any thoughts @aammd?

05-data-structures-part2.Rmd
rmcd1024: On Sat, Aug 29, 2015 at 1:54 PM, Naupaka Zimmerman <notifications@github.com
> wrote:

> In 05-data-structures-part2.Rmd
> <https://github.com/swcarpentry/r-novice-gapminder/pull/59#discussion_r38264827>
> :
>
> >  ```
> >
> >  As you might expect, life expectancy has slowly been increasing over
> >  time, so we see a significant positive association!
> >
> > +> #### Challenge 6 {.challenge}
> > +>
> > +> Write this function: `lm(lifeExp ~ year, data=gapminder)` without
> > +> using the `data` argument. Verify that you get the same answer.
>
> This is a *bad idea* because it will mess up later attempts to use the
> model for prediction. I would take this specific challenge out.
>
>
> It's no problem (and no great loss) to remove this Challenge, but I'm not
understanding the problem:

library(gapminder)
gm1 <- lm(lifeExp ~ year, data=gapminder)
gm2 <- lm(gapminder$lifeExp ~ gapminder$year)
all.equal(predict(gm1), predict(gm2))

So probably I'm missing something...

naupaka: Right, but that is non-standard formula notation and it can get you into trouble (e.g. [here on SO](http://stackoverflow.com/questions/10240017/predict-lm-with-newdata/10242184#10242184)). It will usually work (esp. in simple cases), as you point out, but I think we shouldn't teach it since it could get people into trouble down the road since the names of the things are not what people might expect.

rmcd1024: On Sun, Aug 30, 2015 at 1:30 AM, Naupaka Zimmerman <notifications@github.com
> wrote:

> > +>
>
> > +> Write this function: `lm(lifeExp ~ year, data=gapminder)` without
> > +> using the `data` argument. Verify that you get the same answer.
>
> Right, but that is non-standard formula notation and it can get you into
> trouble (e.g. here on SO
> <http://stackoverflow.com/questions/10240017/predict-lm-with-newdata/10242184#10242184>).
> It will usually work (esp. in simple cases), as you point out, but I think
> we shouldn't teach it since it could get people into trouble down the road
> since the names of the things are not what people might expect.
>

Thank you for the explanation, I've deleted Challenge 6. I hadn't
appreciated the deeper significance of the data argument.

naupaka:
Thanks, @rmcd1024. I'm going to merge.

PR #60, opened 2015-08-30
minor fixes and additions

rmcd1024:
* typos and cleanups
* tip: vector numbering starts at 1
* tip: vector names need not be unique
* added mention of "byrow" option when creating a matrix

### DIFF
--- a/06-data-subsetting.Rmd
+++ b/06-data-subsetting.Rmd
@@ -89,12 +89,15 @@ If we ask for the 0th element, we get an empty vector:
 x[0]
 ```
 
-But what about negative values?
+> ####Vector numbering in R starts at 1 {.callout} 
+> 
+> In many programming languages (C and python, for example), the first
+> element of a vector has an index of 0. In R, the first element is 1.
 
 ### Skipping and removing elements
 
-If we use a negative number, R will return every element *except* for the
-one specified:
+If we use a negative number as the index of a vector, R will return
+every element *except* for the one specified:
 
 ```{r}
 x[-2]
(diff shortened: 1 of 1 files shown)


06-data-subsetting.Rmd
```diff
@@ -210,6 +215,22 @@ x[-which(names(x) %in% c("a", "c"))]
 The `%in%` goes through each element of its left argument, in this case the
 names of `x`, and asks, "Does this element occur in the second argument?".
 
+> #### Tip: Non-unique names {.callout}
+>
+> You should be aware that it is possible for multiple elements in a
+> vector to have the same name. For example:
```
naupaka: Perhaps important to point out here that this is **not** the case for a `data.frame`?

06-data-subsetting.Rmd
```diff
@@ -210,6 +215,22 @@ x[-which(names(x) %in% c("a", "c"))]
...
+
+>```{r}
+> x <- 1:3
+> x
+> names(x) <- c('a', 'a', 'a')  
+> x
+> x['a']
```
naupaka: I suppose the point here is to show that it only pulls out the first matching element of the vector with that name?

rmcd1024: On Sun, Aug 30, 2015 at 1:27 PM, Naupaka Zimmerman <notifications@github.com
> wrote:

> In 06-data-subsetting.Rmd
> <https://github.com/swcarpentry/r-novice-gapminder/pull/60#discussion_r38274347>
> :
>
> > @@ -210,6 +215,22 @@ x[-which(names(x) %in% c("a", "c"))]
> >  The `%in%` goes through each element of its left argument, in this case the
> >  names of `x`, and asks, "Does this element occur in the second argument?".
> >
> > +> #### Tip: Non-unique names {.callout}
> > +>
> > +> You should be aware that it is possible for multiple elements in a
> > +> vector to have the same name. For example:
> > +
> > +>```{r}
> > +> x <- 1:3
> > +> x
> > +> names(x) <- c('a', 'a', 'a')
> > +> x
> > +> x['a']
>
> I suppose the point here is to show that it only pulls out the first
> matching element of the vector with that name?
>
I've tried to clarify the uniqueness discussion.

PR #67, opened 2015-09-28
output of rbind is duplicated

orchid00:
this should be removed
~~~{.output}
Warning in `[<-.factor`(`*tmp*`, ri, value = "g"): invalid factor level, NA
generated

~~~


naupaka:
hi @orchid00 - what happened here? Did the problem end up not being a problem, or?

orchid00:
exactly

PR #70, opened 2015-11-13
Fix format of first tip in RStudio section

karawoo:
The text for the first tip started on the same line as the title and `{.callout}`, causing the tip to not display properly.
screen shot 2015-11-13 at 3 41 44 pm

vs.

screen shot 2015-11-13 at 3 41 51 pm


naupaka:
Thanks!

PR #71, opened 2015-11-19
Update to latest lesson-template

jdblischak:
The most significant change is that the output of code chunks that produce an error or warning will be displayed in red text instead of the standard blue text for code output.

It also replaces the version of `tools/check_knitr_version.R` with the version in lesson-template.

For background on both these changes, see the discussion of the PR that was merged into lesson-template: https://github.com/swcarpentry/lesson-template/pull/284


naupaka:
Thanks @jdblischak. 

So we don't need the `deck.js` support files in there anymore? And why take out the motivation sections?

naupaka:
Ah I see. Rationale was [here](https://github.com/swcarpentry/lesson-template/issues/246) and [here](https://github.com/swcarpentry/lesson-template/issues/264). Looks good to merge. Any other thoughts @aammd?

PR #76, opened 2016-01-06
New challenge regarding subsetting, logical operators and the unique() function

kyrcha:
Exercise for instructor training regarding subsetting using multiple logical operators and introducing the unique() function

### DIFF
--- a/06-data-subsetting.md
+++ b/06-data-subsetting.md
@@ -35,8 +35,8 @@ x
 
 
 ~~~{.output}
-  a   b   c   d   e 
-5.4 6.2 7.1 4.8 7.5 
+  a   b   c   d   e
+5.4 6.2 7.1 4.8 7.5
 
 ~~~
 
@@ -56,8 +56,8 @@ x[1]
 
 
 ~~~{.output}
-  a 
-5.4 
+  a
+5.4
 
 ~~~
 
@@ -69,8 +69,8 @@ x[4]
 
 
 ~~~{.output}
-  d 
-4.8 
+  d
+4.8
 
 ~~~
 
@@ -87,8 +87,8 @@ x[c(1, 3)]
 
 
 ~~~{.output}
-  a   c 
-5.4 7.1 
+  a   c
+5.4 7.1
 
 ~~~
 
@@ -102,8 +102,8 @@ x[1:4]
 
 
 ~~~{.output}
-  a   b   c   d 
-5.4 6.2 7.1 4.8 
+  a   b   c   d
+5.4 6.2 7.1 4.8
 
 ~~~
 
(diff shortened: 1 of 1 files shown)


06-data-subsetting.md
```diff
@@ -1318,3 +1331,29 @@ be changed with the third argument, `drop = FALSE`).
...
+>
+> D. 22
+>
+> Answer: B
+> 
+> The answer can be derived using the following statement
+> ~~~{.r}
+> length(unique(
+>   gapminder[
+>     gapminder$year >= 2002 &
```
naupaka: Why not use `==` here instead of `<=` and `>=`?

06-data-subsetting.md
```diff
@@ -1318,3 +1331,29 @@ be changed with the third argument, `drop = FALSE`).
...
+>
+> Answer: B
+> 
+> The answer can be derived using the following statement
+> ~~~{.r}
+> length(unique(
+>   gapminder[
+>     gapminder$year >= 2002 & 
+>     gapminder$year <= 2007 & 
+>     gapminder$continent=='Asia' &
```
naupaka: I think style guides suggest a space on either side of the `==` in these expressions.

06-data-subsetting.md
```diff
@@ -1318,3 +1331,29 @@ be changed with the third argument, `drop = FALSE`).
...
+> D. 22
+>
+> Answer: B
+> 
+> The answer can be derived using the following statement
+> ~~~{.r}
+> length(unique(
+>   gapminder[
+>     gapminder$year >= 2002 & 
+>     gapminder$year <= 2007 &
```
naupaka: This may be as good a place as any to talk about the differences between `&` vs `&&` and `|` vs `||` in R. From the help docs:

> & and && indicate logical AND and | and || indicate logical OR. The shorter form performs elementwise comparisons in much the same way as arithmetic operators. The longer form evaluates left to right examining only the first element of each vector. Evaluation proceeds only until the result is determined. The longer form is appropriate for programming control-flow and typically preferred in if clauses.

This is often something people (especially those coming from other languages) want to know and get tripped up on.

naupaka:
Hi @kyrcha - thanks for the PR! I made a couple minor suggestions on some of the formatting and minor points on the syntax. Shouldn't take much effort to make the changes.

naupaka:
Thanks again @kyrcha for this contribution -- I think it would be good to rename this pull request with a brief summary of the changes it proposes, instead of the less descriptive "Exercise for instructor training".

kyrcha:
Hi @naupaka - Thank you for your comments. I updated the PR based on your suggestions.

PR #79, opened 2016-01-15
Added challenge question

theavanrossum:
Added challenge question to check learner's understanding of "which" & empty vectors and shows
example of common bug when using "which" to subset


tomwright01:
Excellent addition. Does anyone know why most of this lesson is indented?

naupaka:
@tomwright01 where do you see the indentation? Perhaps I am looking in the wrong place?

tomwright01:
@naupaka I think the problem is that I was looking at the rendered version of the Rmd, which isn't as nice as the rendered md. Also, the new code goes from a challenge section to two tip sections, all of which are indented. Sorry about that, I am happy for this to be merged.

PR #80, opened 2016-01-18
Update names used in RStudio interface

mkuzak:
The Interface in RStudio does not have `Workspace` anymore but `Environment`. Also since some time there is `Viewer` tab in bottom right pane by default.


naupaka:
Looks good. Thanks for the fix.

PR #82, opened 2016-01-22
Add challenge about matrices

fernandomayer:
Proposed challenge for the instructor training wrap up

### DIFF
--- a/04-data-structures-part1.Rmd
+++ b/04-data-structures-part1.Rmd
@@ -67,7 +67,7 @@ typeof(TRUE)
 typeof('banana')
 ```
 
-Note the `L` suffix to insist that a number is an integer. No matter how complicated our analyses become, all data in R is interpreted as one of these basic data types. This strictness has some really important concequences. Go back to your text editor and add add this line to feline-data.csv:
+Note the `L` suffix to insist that a number is an integer. No matter how complicated our analyses become, all data in R is interpreted as one of these basic data types. This strictness has some really important consequences. Go back to your text editor and add add this line to feline-data.csv:
 
 ```{r, eval=FALSE}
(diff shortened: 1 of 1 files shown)


naupaka:
Hi @fernandomayer thanks for the PR. Two things came to mind: 

One is that I think it's easier if we avoid TeX in them when possible, to reduce dependencies, and since the primary output of these documents is the web. 

Second, lines only break if there are multiple (4, I believe) whitespace characters at the end of the line. This is important to get your list of possible challenge answers to display correctly. Alternatively, use markdown ordered list syntax, which requires a period instead of a close parentheses after the number.

fernandomayer:
Hi @naupaka, thanks for the input.

I've made a second commit (f07a38) removing the TeX input (using an R output instead), and fixed the markdown list by using period instead of parenthesis. (The list with parenthesis I've copied from the first challenge).

Hope this renders properly now.

Thanks.

naupaka:
Thanks for the fixed @fernandomayer. We have had some additional changes to the same lesson you changed since you initially submitted your PR. Would you mind updating your changes based on the current version of the lesson? And, while you're at it, could you fix the list for Challenge 1? The ordered with with parentheses doesn't render properly (although the list from the challenge you added now looks fine, thanks for fixing that).

fernandomayer:
Hi @naupaka 

I've merged the new changes and kept my challenge at the end of the lesson.

I've also fixed two small typos that I saw while merging.

I didn't see any list in Challenge 1 in this revision, so I'm not sure what I was supposed to do about it.

Anyway, let me know if anything else is needed here.

Thanks.

naupaka:
That must have been on a prior version of this lesson. In any case, thanks for the fixes, and for the PR!

PR #87, opened 2016-01-24
update 06-subsetting on logical operations

davharris:
I made the following changes to the callout on logical operators:

* In the title, "chain" becomes "combine": my impression is that the word "chain" is not commonly used as a verb by non-programmers

* In the title, "operations" becomes "conditions": this section mostly talks about combining the `TRUE`s and `FALSE`s themselves, rather than the operations such as `>` that produced them.

* Add an example for why we'd want to combine multiple logical vectors

* Fix an incorrect statement regarding `&&` and `||`'s recycling behavior

* Add logical negation, `all`, and `any`.

If the section is too long now, I'd suggest dropping the double operators (`&&` and `||`) first, then `all` and `any` if needed..

### DIFF
--- a/06-data-subsetting.Rmd
+++ b/06-data-subsetting.Rmd
@@ -330,7 +330,7 @@ x[x > 7]
 > ## Tip: Combining logical conditions {.callout}
 >
 > There are many situations in which you will wish to combine multiple logical
-> criteria For example, we might want to find all the countries that are
+> criteria. For example, we might want to find all the countries that are
 > located in Asia **or** Europe **and** have life expectancies within a certain
 > range. Several operations for combining logical vectors exist in R:
 >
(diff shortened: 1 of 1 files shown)


06-data-subsetting.Rmd
```diff
@@ -327,17 +327,37 @@ use them to succinctly subset vectors:
 x[x > 7]
 ```
 
-> ## Tip: Chaining logical operations {.callout}
+> ## Tip: Combining logical conditions {.callout}
 >
-> There are many situations in which you will wish to combine multiple conditions.
-> To do so several logical operations exist in R:
+> There are many situations in which you will wish to combine multiple logical
+> criteria For example, we might want to find all the countries that are
```
naupaka: Missing period after 'criteria'

naupaka:
Thanks for the PR @davharris. I'm good to merge after you fix the minor typo. I think your changes help to clarify the callout content.

davharris:
Thanks.  Added the period.

PR #89, opened 2016-02-06
reimagining data types and data structures

BillMills:
Hi folks - this PR remixes sections 4 and 5 (intro to data types and data structures), following the pattern tested out by @lmguzman [at UBC](http://lmguzman.github.io/2016-01-29-UBC/index.html) and myself  [at ResBaz SFU](http://bgran.de/2016-02-02-SFU/) over the past week; the core motivation is to **introduce types and structures by exploring the data.frame**.

Understanding types and structures is crucial for successfully wielding R - the problem being, that beginners really struggle to see why this is so, and are often turned off by how dry the topic seems. Instead of marching them through a gallery of types and structures, this method starts by loading a csv right away, and performing common manipulations which quickly lead to illustrations and conversations on why data types exist, what the properties of data structures are, and how they inform best practice for looking at and understanding data. Feedback was positive from students on both runs at this version.

@aammd @theavanrossum @cloucks have all seen this taught live - did I miss anything / anything we should change in hindsight? If someone better at knittr than I would jam this through that framework to doublecheck it too, I'd be much obliged. Feedback welcome, let me know.

### DIFF
--- a/04-data-structures-part1.Rmd
+++ b/04-data-structures-part1.Rmd
@@ -32,8 +32,6 @@ cats <- read.csv(file="data/feline-data.csv")
 cats
 ```
 
-The extra arguments after the filename help R get its hands around our data; `header=TRUE` lets it know there's a title row, and `sep=","` lets it know our columns are separated by `,` characters. Note that if you like to maintain the English idiom and put spaces after your commas, you'll need to use `sep=", "`; either is fine, as long as you're consistent.
-
 We can begin exploring our dataset right away, pulling out columns and rows via the following:
 
 ```{r}
--- a/04-data-structures-part1.Rmd
+++ b/04-data-structures-part1.Rmd
@@ -159,7 +159,7 @@ z <- as.logical(y)
 z
 ```
 
-As you can see, some surprising things can happen when R forces one basic data type into another! Nitty-gritty of type coercion aside, the point is: if you're data doesn't look like what you thought it was going to look like, type coercion may well be to blame; make sure everything is the same type in your vectors and your columns of data.frames, or you will get nasty surprises!
+As you can see, some surprising things can happen when R forces one basic data type into another! Nitty-gritty of type coercion aside, the point is: if your data doesn't look like what you thought it was going to look like, type coercion may well be to blame; make sure everything is the same type in your vectors and your columns of data.frames, or you will get nasty surprises!
 
 Concatenate will also append things to an existing vector:
 
@@ -277,7 +277,7 @@ We can now understand something a bit surprising in our data.frame; what happens
 typeof(cats)
 ```
 
-We see that data.frames look like lists 'under the hood' - this is because a data.frame is really a list of vectors, as they have to be - in order to hold those columns that are a mix of vectors and factors, the data.frame needs something a bit more flexible than a vector to put all the columns together into a familiar table.
+We see that data.frames look like lists 'under the hood' - this is because a data.frame is really a list of vectors and factors, as they have to be - in order to hold those columns that are a mix of vectors and factors, the data.frame needs something a bit more flexible than a vector to put all the columns together into a familiar table.
 
 ## Matrices
 
@@ -345,7 +345,7 @@ x
 > a long way to making our lives easier in R. 
 >
 
-> ## Challenge 1 {.challenge}
+> ## Solution to Challenge 1 {.challenge}
 > ```{r}
 > x <- [11:20]
 > firstFive <- head(x, 5)
@@ -354,7 +354,7 @@ x
 > ```
 >
 
-> ## Challenge 2 {.challenge}
+> ## Solution to Challenge 2 {.challenge}
 > ```{r}
 > cats <- read.table(file="data/feline-data.csv", header=TRUE, sep=",", stringsAsFactors=FALSE)
 > str(cats$coat)
@@ -370,7 +370,7 @@ x
 > `length(x)`?
 >
 > ```{r}
-> x <- matrix(rnorm(18), ncol=6, nrow=3)
+>x <- matrix(0, ncol=6, nrow=3)
 > length(x)
 > ```
 > 
@@ -394,7 +394,7 @@ x
 > ```
 >
 
-> ## Challenge 5 {.challenge}
+> ## Solution to Challenge 5 {.challenge}
 > ```{r}
 > dataTypes <- c('double', 'complex', 'integer', 'character', 'logical')
 > dataStructures <- c('data.frame', 'vector', 'factor', 'list', 'matrix')
--- a/04-data-structures-part1.Rmd
+++ b/04-data-structures-part1.Rmd
@@ -9,7 +9,6 @@ minutes: 90
 source("tools/chunk-options.R")
 ```
 
-
 > ## Learning Objectives {.objectives}
 >
 > - To be aware of the different types of data
@@ -192,7 +191,7 @@ length(x)
 Finally, you can give names to elements in your vector, and ask for them that way:
 
 ```{r}
-x <- 1:4
+x <- 5:8
 names(x) <- c("a", "b", "c", "d")
 x
 x['b']
@@ -244,7 +243,7 @@ typeof(CATegories[1])
 > ## Challenge 2 {.challenge}
 > When we loaded our `cats` data, the coats column was interpreted as a factor; 
 > try using the help for `read.table` to figure out how to keep text columns as character vectors instead of factors; 
-> then write a command or two to show that coats actually is a character vector when loaded in this way.
+> then write a command or two to show that the cats$coats column actually is a character vector when loaded in this way.
 >
 
 In modeling functions, it's important to know what the baseline levels are. This is assumed to be the
--- a/04-data-structures-part1.Rmd
+++ b/04-data-structures-part1.Rmd
@@ -2,7 +2,7 @@
 layout: page
 title: R for reproducible scientific analysis
 subtitle: Data structures
-minutes: 45
+minutes: 90
 ---
 
 ```{r, include=FALSE}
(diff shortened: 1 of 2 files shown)


rmflight:
This looks really nice @BillMills, and I think is a natural way to introduce types using some real data, and the pitfalls of working with it.

After reading through it I can't see anything missed, other than logicals. But I'm not sure how you would introduce that easily, except maybe by adding another species, or male/female perhaps?

05-data-structures-part2.Rmd
```diff
@@ -1,245 +1,132 @@
...
 > ## Learning Objectives {.objectives}
 >
-> * Become familiar with data frames
-> * Be able to read regular data into R
+> - To learn how to manipulate a data.frame in memory
+> - To tour some best practices of exploring and understanding a data.frame when it is first loaded.
 >
 
-## Data frames
+At this ponit, you've see it all - in the last lesson, we toured all the basic data types and data structures in R. Everything you do will be a manipulation of those tools. But a whole lot of the time, the star of the show is going to be the data.frame - that table that we started with that information from a CSV gets dumped into when we load it. In this lesson, we'll learn a few more things about working with data.frame.
```
naupaka: type: ponit -> point

And should data.frame be `data.frame`? Not sure if there is a SWC lesson style guide that includes such things.

05-data-structures-part2.Rmd
```diff
@@ -1,245 +1,132 @@
...
-data frames are really lists, where each element is an atomic vector,
-with the added restriction that they're all the same length.  As you
-will see, if we pull out one column of a data frame,we will have a
-vector. You will probably find that data frames are more complicated
-than vectors and other data structures we have considered, but they
-provide powerful capabilities.
-
-
-Data frames can be created manually with the `data.frame` function:
+We learned last time that the columns in a data.frame were vectors, so that our data is reliable and consistent throughout the column. As such, if we want to add a new column, we need to start by making a new vector:
```
naupaka: pedantic grammar Q: data is vs. data are vs data set is

Is there a SWC position on this issue?

naupaka:
@BillMills this is quite a PR! I much prefer the new approach, it seems more easily graspable and I think it works better flow-wise as well. I had a couple very minor in-line comments, but won't have a chance to do a detailed line by line proofing check for the next little while. If anyone spots any other typos or otherwise, please add them in comments. Overall though, I am definitely +1 to merge this in soon.

05-data-structures-part2.Rmd
tomwright01: I'd say data.frame is a name, `data.frame()` is the function.

05-data-structures-part2.Rmd
tomwright01: Not sure the data is reliable just because it's a vector. Wish it was though!

05-data-structures-part2.Rmd
```diff
@@ -1,245 +1,132 @@
...
-levels. Why? Let's examine the first column.  We can access a column
-in a `data.frame` by using the `$` operator.
-
+Another thing to look out for has emerged - when R creates a factor, it only allows whatever is originally there when our data was first loaded, which was 'black', 'calico' and 'tabby' in our case. Anything new that doesn't fit into one of its categories is rejected as nonsense, until we explicitly add that as a *level* in the factor:
 
 ```{r}
-class(df$id)
-str(df)
+levels(cats$coat)
+levels(cats$coat) <- c(levels(cats$coat), 'tortoise')
```
tomwright01: Pedantic, where I come from tortoise is an animal, tortoiseshell is the colour. Happy to be corrected if that's not an international convention.

04-data-structures-part1.Rmd
```diff
@@ -2,336 +2,256 @@
...
-> ## Tip: "Character Vectors" {.callout}
->
-> You will sometimes hear the term "character vector", especially in
-> warning or error messages. This is a somewhat confusing and unfortunate
-> name. Remember that the type "character" really means some text
-> wrapped in quotation symbols.
->
+```{r, eval=FALSE}
+tabby,2.3 or 2.4,TRUE
+```
```
tomwright01: This is a little confusing as it requires editing the csv file. Do you do this within RStudio or in an external editor? I like the idea of doing the editing in a spreadsheet and using the `file-->save as` function. Perhaps this should be made more explicit.

04-data-structures-part1.Rmd
```diff
@@ -2,336 +2,256 @@
...
 x
 ```
 
-Or we can use the concatenate function to combine any values we like into
-a vector (so long as they're the same atomic type!).
+Just like you might be familiar with from vectors elsewhere, a vector in R is essentially an ordered list of things, with the special condition that *everything in the vector must be the same basic data type*. If you don't choose the datatype, it'll default to `logical`; or, you can declare an empty vector of whatever type you like.
 
 ```{r}
-x <- c(10, 12, 45, 33)
+x <- vector('character', length=3)
```
tomwright01: I prefer to be explicit x<-vector(mode='character',length=3)

04-data-structures-part1.Rmd
```diff
@@ -2,336 +2,256 @@
...
+## Factors
 
-We can create one with the `matrix` function. Let's generate some random data:
+We said that columns in data.frames were vectors:
 
 ```{r}
-set.seed(1) # make sure the random numbers are the same for each run
-x <- matrix(rnorm(18), ncol=6, nrow=3)
-x
+str(cats$weight)
```
tomwright01: cats$weight is currently type factor (on my system).

04-data-structures-part1.Rmd
```diff
@@ -339,125 +259,110 @@ In this case, we've explicitly told R that "control" should represented by 1, an
...
+> ## Solution to Challenge 1 {.challenge}
 > ```{r}
-> xx <- c("a", TRUE)
-> xx
-> typeof(xx)
+> x <- [11:20]
+> firstFive <- head(x, 5)
+> thirdToFifth <- tail(firstFive, 3)
+> names(thirdToFifth) <- c('S', 'W', 'C')
 > ```
```
tomwright01: This is a strange way to answer this challenge, since square brackets and sequences have been introduced `x[3:5]` works just as well and could be added as an alternate answer.

05-data-structures-part2.Rmd
```diff
@@ -1,245 +1,132 @@
...
-and last columns into factors, not character vectors. There are no
-pre-existing levels "g" and "G", so the attempt to add these values
-fails. A row was added to the data frame, only there are missing
-values in the factor columns:
+Alternatively, we can change a factor column to a character vector; we lose the handy categories of the factor, but can subsequently add any word we want to the column without babysitting the factor levels:
 
 ```{r}
-df
+str(cats)
+cats$coat <- as.character(cats$coat)
```
tomwright01: I would prefer to use
```{r}
cats$coat <- levels(cats$coat)[cats$coat]
```
as.character() hides the problem with factors containing numeric values.
```{r}
x <- as.factor(c(2,4,6))
as.numeric(x)
as.numeric(levels(x)[x])
```

05-data-structures-part2.Rmd
```diff
@@ -1,245 +1,132 @@
...
-
-There are multiple ways to delete a row containing `NA`:
+In either case, we need to reassign our variable to persist the changes:
 
 ```{r}
-df[-7, ]  # The minus sign tells R to delete the row
-df[complete.cases(df), ]  # Return `TRUE` when no missing values
-na.omit(df)  # Another function for the same purpose
-df <- na.omit(df)
+cats = na.omit(cats)
```
tomwright01: cats <- na.omit(cats)

tomwright01:
I also really like this PR. Need to decide how to create the original csv file. I'm in favour of explicitly using an external editor rather than trying to package it with the gapmider data. That would highlight the problem with data entry into spreadsheets "2.1 or 2.2", this is a very common problem with data I receive "value not in chart", "can't read hand writing". 
Otherwise I'm looking forward to using this material in a workshop.

sritchie73:
I haven't had a chance to go through the PR, but I really like this idea, and have been wanting to refine the materials in this direction for some time. I taught the lesson materials following a similar format at the novice R workshop I ran for Melbourne ResBaz last week.

05-data-structures-part2.Rmd
BillMills: happy with either, let me know what you'd prefer.

04-data-structures-part1.Rmd
BillMills: Thanks for all your feedback, Tom! I'm open-minded on this one; I chose to be intentionally mum to let instructors do as they thought best here, but if we'd like to recommend a particular method, that's fine be me.

04-data-structures-part1.Rmd
BillMills: Good catch! I added a note to explicitly change it back above.

04-data-structures-part1.Rmd
```diff
@@ -2,336 +2,271 @@
...
 
-* vector
-* factor
-* list
-* matrix
-* data.frame
+## Data Types
 
-For now, let's focus on vectors in more detail, to discover more about data types.
+If you guessed that the last command will return an error because 2.1 plus black is nonsense, you're right - and you already have some intuition for an important concept in programming called *data types*. We can ask what type of data something is:
```
aammd: oh i :heart: this part

04-data-structures-part1.Rmd
```diff
@@ -2,336 +2,271 @@
...
-    * double (i.e. decimal) (e.g, `-24.57`, `2.0`, `pi`)
-* complex (i.e. complex numbers) (e.g, `1 + 0i`, `1 + 4i`)
-* text (called "character" in R) (e.g, `"a"`, `"swc"`, `'This is a cat'`)
+```{r}
+cats <- read.table(file="data/feline-data.csv", header=TRUE, sep=",")
+cats
+```
 
-There are a few functions we can use to interrogate data in R to determine its type:
+The extra arguments after the filename help R get its hands around our data; `header=TRUE` lets it know there's a title row, and `sep=","` lets it know our columns are separated by `,` characters. Note that if you like to maintain the English idiom and put spaces after your commas, you'll need to use `sep=", "`; either is fine, as long as you're consistent.
```
aammd: From the way I was taught, I feel like `read.csv` is a more common function. It sets the `header` and `sep` values automatically.  So that could be replaced, and much of L35 removed.
That said, perhaps part of the reason for keeping `read.table` in here is to have this discussion of its arguments? If this were a first introduction to functions, I think that would make sense. On the other hand, if the goal here is to get students working with data ASAP then best to just use `read.csv`

04-data-structures-part1.Rmd
```diff
@@ -2,336 +2,271 @@
...
+typeof(TRUE)
+typeof('banana')
+```
 
-* logical (e.g., `TRUE`, `FALSE`)
-* integer (e.g., `2L`, `as.integer(3)`)
-* numeric (real or decimal) (e.g, `2`, `2.0`, `pi`)
-* complex (e.g, `1 + 0i`, `1 + 4i`)
-* character (e.g, `"a"`, `"swc"`)
+Note the `L` suffix to insist that a number is an integer. No matter how complicated our analyses become, all data in R is interpreted as one of these basic data types. This strictness has some really important concequences. Try adding this row to your cat data:
```
aammd: Perhaps something like: "Try adding this row" --> "open a text editor and add this row" or maybe "go back to your text editor and add.."

04-data-structures-part1.Rmd
aammd: I'm with @BillMills -- instructors could probably choose. That said, my instinct is that spreadsheets are trouble. Every student will have different spreadsheets, and they may not know how to save to csv. Not to mention confusion arising from differences in "smart quotes" across different computers. 
I'd vote for writing the csv file directly in the Rstudio text editor.

tomwright01: Perhaps just change line 72 to "Use an external editor (e.g. nano or excel) to add this line to the cat data."

04-data-structures-part1.Rmd
```diff
@@ -2,336 +2,271 @@
...
+Concatenate will also append things to an existing vector:
 
 ```{r}
-x <- c("a", "b", "c")
-as.numeric(x)
-as.logical(x)
+x <- c('a', 'b')
+x
+x <- c(x, 'SWC')
+x
```
aammd: Is it worth adding a callout here to the for-loop section? Just to point out that while appending things to vectors in this way will work, it is a bad practice. (because it slows R down so much)

Additionally, I think it's tricky in a classroom setting to overwrite `x` like this.  Better to suggest instructors do 
````r
x <- c(x, 'SWC')
```

Or even give the vectors memorable names, like `first_two_letters` and `first_two_swc` or whatever

04-data-structures-part1.Rmd
```diff
@@ -339,125 +274,109 @@ In this case, we've explicitly told R that "control" should represented by 1, an
...
-different kinds of results into a single object that a function can return. In
-fact many R functions which return complex output store their results in a list.
+and we can ask for and put values in the elements of our matrix with a couple of different notations:
 
-## Challenge solutions
+```{r}
+x[1,1] = 1
+x
+x[1][1]
+x[1][1] = 2
```
aammd: let's be consistent and use `<-` for assignment

04-data-structures-part1.Rmd
```diff
@@ -2,336 +2,271 @@
...
+typeof(coats[1])
+typeof(CATegories[1])
 ```
 
-This reveals something important: while factors look (and often behave) like
-character vectors, they are actually integers under the hood, and here, we can
-see that "no" is represented by a 1, and "yes" a 2.
+> ## Challenge 2 {.challenge}
+> When we loaded our `cats` data, the coats column was interpreted as a factor; 
+> try using the help for `read.table` to figure out how to keep text columns as character vectors instead of factors;
```
aammd: should this be `?read.table` or is part of the exercise to help them remember how to do that.

04-data-structures-part1.Rmd
```diff
@@ -2,336 +2,271 @@
...
-str(x)
+typeof(coats[1])
+typeof(CATegories[1])
 ```
 
-This reveals something important: while factors look (and often behave) like
-character vectors, they are actually integers under the hood, and here, we can
-see that "no" is represented by a 1, and "yes" a 2.
+> ## Challenge 2 {.challenge}
+> When we loaded our `cats` data, the coats column was interpreted as a factor;
```
aammd: Do they know this already? Perhaps this should be a question ("what is the type of the `coats` column?" or even "Is there a factor in `cats`? what is its name?")

04-data-structures-part1.Rmd
```diff
@@ -2,336 +2,271 @@
...
 ```
 
-This reveals something important: while factors look (and often behave) like
-character vectors, they are actually integers under the hood, and here, we can
-see that "no" is represented by a 1, and "yes" a 2.
+> ## Challenge 2 {.challenge}
+> When we loaded our `cats` data, the coats column was interpreted as a factor; 
+> try using the help for `read.table` to figure out how to keep text columns as character vectors instead of factors; 
+> then write a command or two to show that the cats$coats column actually is a character vector when loaded in this way.
+>
```
aammd: `cats$coats`

04-data-structures-part1.Rmd
```diff
@@ -2,336 +2,271 @@
...
-Vectors can be named:
+Finally, you can give names to elements in your vector, and ask for them that way:
 
 ```{r}
-x <- 1:4
+x <- 5:8
 names(x) <- c("a", "b", "c", "d")
 x
+x['b']
 ```
```
aammd: As much as I love this form, I might be better to go with `x$b`. This would provide a nice callback to L40, where we talk about `cats$weight`. i.e. here is another chance to explore what the `$` operator does

05-data-structures-part2.Rmd
```diff
@@ -1,245 +1,132 @@
...
-and last columns into factors, not character vectors. There are no
-pre-existing levels "g" and "G", so the attempt to add these values
-fails. A row was added to the data frame, only there are missing
-values in the factor columns:
+Alternatively, we can change a factor column to a character vector; we lose the handy categories of the factor, but can subsequently add any word we want to the column without babysitting the factor levels:
 
 ```{r}
-df
+str(cats)
+cats$coat <- levels(cats$coat)[cats$coat]
```
aammd: This is elegant in the extreme, and nicely illustrates why the secret numeric nature of factors is a feature, not a bug. 
However wouldn't it be clearer to use `as.character`, since we've already introduced the gang of coercion functions?

aammd:
This is a really awesome PR which I think really improves this lesson. Well done @BillMills !

This is perhaps the trickiest module in the entire lesson, because here students will be meeting three things for the first time (afaik):

* functions
* subsetting
* the five basic types of data, and _also_ vectors, lists, matrices and data.frames

The lesson focusses on the third point above, and references the first two only implicitly (they each have their own modules later). I think it would be worthwhile to allow a slight digression for each, just to let people know that there is more on the way ("`read.table` is an example of a _function_, and has arguments separated by commas.." "`cats$weight` gives us the part of `cats` that has the label `weight`. we'll learn more about how to get only part of an object later (this is called subsetting)"  
That said, I think we should remove as much of the variation in subsetting as we can (e.g. perhaps take out the bit about dropping rows of a matrix.)

on a related point: I have my doubts about `rbind` and `cbind`. I realize I may be in the minority view here. I just can't imagine many cases where combining an independent row or column to a data.frame is the best practice. People have to use extreme caution to make sure that everything is where it is supposed to be (e.g. that a vector is in the exact correct sequence to be lined up with a dataframe via `cbind`). I just fear that learners may not realize the danger. 

However, modifying data and adding the new data as a column is a really important operation, so we should show it _somehow_. How about something like this: 

```r
## lets calculate the mass of our cats in kg
cats$weight_in_kg <- cats$weight / 2.2
```

04-data-structures-part1.Rmd
BillMills: I knew it! I knew I'd forget *somewhere* :)

04-data-structures-part1.Rmd
BillMills:
```
names_example <- 5:8
names(names_example) <- c("a", "b", "c", "d")
names_example
names_example$b
```
`Error in names_example$b : $ operator is invalid for atomic vectors`

04-data-structures-part1.Rmd
BillMills: Done, remaining coy about precisely which editor to use - this is definitely clearer, but I share @aammd's concerns over a round trip through a spreadsheet; I'll leave it to the instructor to lead.

05-data-structures-part2.Rmd
BillMills: happy to defer to the R experts here - I think @aammd asked for `as.character`, but I'm happy with any consensus.

BillMills:
Thanks all for your comments! @aammd, a few responses:

>  I think it would be worthwhile to allow a slight digression for each, just to let people know that there is more on the way

Forward references to material-to-come are generally frowned upon in the unwritten SWC guide to pedagogical style; it bloats people's working memory with a bunch of things they don't really have a mental model of yet. Instead, pointing back to tie everything together is usually 'the way' - but if there's a specific reason why that's a disservice here, I'm game.

> remove as much of the variation in subsetting as we can

and

> on a related point: I have my doubts about rbind and cbind

I kept these in not for any strong opinion, but in sheer deference to the original authors of the lesson who know more R than me - happy to cut anything people see fit - does anyone else want to weigh in, or should we trim these things exactly per @aammd's suggestion?

ateucher:
Over the last two days I taught an R workshop at work (not SWC), largely using this material, including @BillMills changes proposed here. Overall I really liked the lesson! When I find some time over the next few days I'll add some comments to the discussion here :)

04-data-structures-part1.Rmd
aammd: WHOOPS! nevermind! :cold_sweat: :sweat_smile:

aammd:
@BillMills 

> Forward references to material-to-come are generally frowned upon in the unwritten SWC guide to pedagogical style; it bloats people's working memory with a bunch of things they don't really have a mental model of yet.

ah that makes sense! Well, perhaps one of those little "tip" boxes would be good, just to go into more detail about functions and subsetting? A shallow introduction now, and then later lessons can link back to these. That said, this probably doesn't belong in this PR

@ateucher looking forward to hearing what you think also! After Andy's comments, perhaps we can get a final comment/approval from everyone else involved so far ( @tomwright01 @naupaka and @sritchie73 & @rmflight  ) before merging?

ateucher:
Hi all, sorry for taking so long to get back to you with my comments.

First of all, I REALLY like this lesson. The flow is great, and it introduces concepts in a context in which students can see the applicability immediately (I actually got this feedback directly from a couple of students).

I agree with most of @aammd 's comments, and I'll summarize how I addressed them, as well as a few other modifications I made (which I realize are not necessarily about this PR).

1. I introduced functions (using, not creating) in the introduction lesson, so this isn't their first encounter with them. I talked about what a function is, how to call it, what is an argument, default argument values, what is a return value, etc. in that earlier lesson.

2. I removed all square-bracket subsetting from my version of this lesson, and really focused on using this to introduce data types and structures. I actually did subsetting immediately AFTER plotting, which worked pretty well, because during the plotting section I had questions like 'What if I want to plot just the African countries?'.

3. I agree 100% with @aammd 's comments about `rbind` and `cbind`. I removed all references to both, and replaced the `cbind` with creating a new column like Andrew's suggestion (though I think those weights must be in kg already - if they're in lbs, they must be really small cats!).

4. **Creating the "cats" csv:** In my organization, Excel is a religion, so I didn't really need to worry about people having different software. I got participants to create the cats data in Excel to illustrate *how to save an Excel file as a csv*. I then had them open the file in a text editor to show them exactly what a csv is - just a text file with columns delimted by commas. I explained that while R can read Excel files directly, it much prefers delimited text and you are less likely to run into problems if you do it this way.  

5. I replaced `read.table()` with `read.csv()` since csvs are what they will encounter most of the time. I mentioned that tab-delimited files are also very common, and that `read.delim()` deals well with those, and that if they have a file delimited by something other than commas or tabs, then `read.table()` is available, and more flexible.

6. When talking about data types, I also talked about using `class()` to find out what an object is - this resonated especially well when explaining that factors are integers under the hood - `class(myFactor)` returns `"factor"`, while `typeof(myFactor)` returns `"integer"`. Also, when you use `str()` on a data.frame, it is the class of each column that is displayed, not the atomic type.  I know when I was learning R the distinction between atomic types and classes really confused me for a while.

7. In the initial creation of the "cats " csv, I had them encode `likes_string` as `0` or `1` rather than `TRUE` or `FALSE`. Then I used `as.logical(cats$likes_string)` to illustrate type coercion in a context that they are likely to come across at some point.

05-data-structures-part2.Rmd
```diff
@@ -1,245 +1,134 @@
...
-
-## Adding rows to a data frame 
-
-To add a row we use `rbind`:
+Now how about adding rows - in this case, we saw last time that the rows of a data.frame are made of lists:
 
 ```{r}
-df <- rbind(df, list("g", 11, 42, 0, "G"))
+newRow <- list("tortoiseshell", 3.3, TRUE, 9)
+cats <- rbind(cats, newRow)
```
ateucher: Using `rbind` to append a row as a `list` to illustrate (one of) the pitfalls of factors is a bit contrived since if you constructed the new row using `data.frame` instead of `list`, R will automatically expand the factor levels to include the new one. eg:

```r
newRow <- data.frame(coat = "tortoiseshell", weight = 3.3, likes_string = TRUE, age = 9)
cats <- rbind(cats, newRow)
```

aammd: Good point @ateucher . but either way, it sounds like you didn't teach `rbind` or `cbind` at all?

aammd:
@ateucher I love all your points, particularly 5 and 7. I think they would make excellent refinements to this PR (which is already great)

@naupaka @tomwright01 what shall we do? wouldn't it be best to get @BillMills 's changes in asap, so that @ateucher could PR his own refinements?

05-data-structures-part2.Rmd
ateucher: Yup, that's true

naupaka:
Yes I agree. Think @BillMills addressed all the review comments? If so, I am good to merge this one and then perhaps let @ateucher add his tweaks as a new PR?

aammd:
d'accord. Let's hit that Big Green Button :green_heart:

ateucher:
@aammd @naupaka @tomwright01 I'll package up my suggestions as a few smaller PRs - that way you can decide what you want in/out at a more granular level.

PR #91, opened 2016-02-21
fix legend label error

atyre2:
The example code under *Modifying Text* uses `scale_fill_discrete()` to change the label on the legend from "continent" to "Continent". However this doesn't work, because it is a color scale. The fix is to change to `scale_colour_discrete(name="Continent")`.


naupaka:
Good catch. Thanks for the PR, @atyre2

PR #92, opened 2016-02-21
Changed solution to Challenge 2 to use ggplot()

atyre2:
The solutions to challege 2 in the vectorisation page use base graphics, but the visualization page shows learners `ggplot()`. I updated the Rmd to use `ggplot()` to produce the solution.


naupaka:
Good catch @atyre2. That's the only spot in all the lessons that uses base R plotting, and it comes after the ggplot lesson, so it makes sense to use `ggplot()` in this case. I'm happy to merge. Any other thoughts @aammd @tomwright01?

PR #95, opened 2016-03-09
Remove square-bracket subsetting from data lessons

ateucher:
As per the discussions in #89 , here is a pull request containing changes that remove square-bracket subsetting from the data lessons (04 and 05).

In addition, I:
- replaced `read.table` with `read.csv`, 
- added a bit of an introduction to classes (very basic), 
- made `cats$likes_string` with 1s and 0s instead of `TRUE` and `FALSE` to help illustrate coercion.

Things I didn't do:
- edit the introduction lesson to introduce function usage
- get rid of `rbind` and `cbind`
- rework subsetting lesson. I think it still works as is, but it might be worth having another look at it, and potentially incorporate moar cats.

I'm happy to work on those things, but I don't think it will happen in the very near future. If this PR looks okay, I think the overall R lesson will still work (though it would probably be worth a closer look) ... then I (or somebody else) can do those things if they are deemed necessary.

### DIFF
--- a/04-data-structures-part1.Rmd
+++ b/04-data-structures-part1.Rmd
@@ -9,10 +9,13 @@ minutes: 90
 source("tools/chunk-options.R")
 
 ## Save the data
-cats <- data.frame(coat = c("calico", "black", "tabby"),
+cats_orig <- data.frame(coat = c("calico", "black", "tabby"),
                    weight = c(2.1, 5.0, 3.2), 
                    likes_string = c(1, 0, 1))
-write.csv(cats, "data/feline-data.csv", row.names = FALSE, quote = FALSE)
+cats_bad <- data.frame(coat = c("calico", "black", "tabby", "tabby"),
+                   weight = c(2.1, 5.0, 3.2, '2.3 or 2.4'), 
+                   likes_string = c(1, 0, 1, 1))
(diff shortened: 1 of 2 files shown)


naupaka:
@ateucher thanks for the PR!  We just merged some pretty substantial changes to the data type lessons, and now there are a bunch of merge conflicts with your PR. If you can pull the current version and add your changes on top of that, I'd be happy to merge this in. It may also be that some of your changes are now redundant with the improvements to the lessons. Or if you think the new versions of the lessons are OK and additional changes are unnecessary, then I will close this PR.

aammd:
@ateucher @naupaka I'm going to re-examine this for merge conflicts with the most recent version of our lessons (i.e. a9dc3d6 )

ateucher:
Yeah, sorry about that - I should have dealt with this then, as I'm sure there are many more changes since that have introduced even more conflicts...

ateucher:
@aammd @naupaka @tomwright01 Since there are so many other issues being fixed here today, I wonder if throwing this big one into the mix might make things a bit more difficult to track. Proposal: After the bug BBQ (which unfortunately I can't take part in today), I will revisit my PR, address any conflicts, then resubmit?

Issue #62, opened 2015-09-07
Links are out of date in reference.md


naupaka:
The content and links in `reference.md` need to be updated to match the new lesson order and names.

naupaka:
Some additional errors from `make check` on `reference.md` that need fixing:
~~~
ERROR: In ./reference.md: Document is missing expected heading: Glossary
ERROR: In ./reference.md: The linked page 05-data-structures-part2.html exists, but the link text 'Reading data' does not match the (sub)title of that page, 'Exploring Data Frames'.
ERROR: In ./reference.md: The linked page 07-functions.html exists, but the link text 'Functions' does not match the (sub)title of that page, 'Creating functions'.
ERROR: In ./reference.md: The linked page 08-plot-ggplot2.html exists, but the link text 'Creating graphics' does not match the (sub)title of that page, 'Creating publication quality graphics'.
ERROR: In ./reference.md: The linked page 15-wrap-up.html exists, but the link text 'Defensive Programming' does not match the (sub)title of that page, 'Wrapping up'.
~~~

tomwright01:
Opened new issue for missing glossary #158

PR #97, opened 2016-03-22
Updated reference.md to reflect new lesson structure

tomwright01:
Issue #62 
I kept a lot of the original information on data types and structures although the new data structure material presents it in a different format as I think it makes a great summary resource.


naupaka:
Thanks @tomwright01! I'm keen to merge this ASAP, not only because it keeps things up to date, but it also will fix a whole bunch of the errors that pop up during `make preview`.  I'll wait a couple days in case there is any other feedback, and then merge.

PR #98, opened 2016-03-24
Add solution for challenge 2

jcoliver:


naupaka:
Thanks for the fix! Want to take a crack at a solution for Challenge 3 in this lesson too?

PR #101, opened 2016-04-12
Update 07-functions.Rmd

orchid00:
Modified challenge 3, which uses paste. It is not standard to use collapse, collapse is an optional parameter in that line 293 the attribute sep should be used instead. Hence I also modified the Note on line 303.

### DIFF
--- a/07-functions.Rmd
+++ b/07-functions.Rmd
@@ -290,7 +290,7 @@ which is much better than in our first attempt where we just got a vector of num
 >
 > ```{r}
 > best_practice <- c("Write", "programs", "for", "people", "not", "computers")
-> paste(best_practice, collapse=" ")
+> paste(best_practice, sep=" ")
 > ```
 >
 >  Write a function called `fence` that takes two vectors as arguments, called
@@ -300,8 +300,8 @@ which is much better than in our first attempt where we just got a vector of num
 > fence(text=best_practice, wrapper="***")
 > ```
 >
-> *Note:* the `paste` function has an argument called `sep`, which specifies the
-> separator between text. The default is a space: " ". The default for `paste0`
+> *Note:* the `paste` function has an argument called `collapse`, which is equivalent 
+> to use sep = "" (add strings without a separator). The default is a space: " ". The default for `paste0`
 > is no space "".
 >
 


naupaka:
Hi @orchid00 thanks for the PR. `sep` and `collapse` [are actually used for different things](https://gist.github.com/briandk/d9231ba1e2603eed0df1), but they produce the same result when used  with a single vector, as in this case. Since most people use `paste()` to combine elements of a single vector instead of multiple vectors, most people generally just use `sep`, as you point out. The difference between the two only really comes into play when you are combining two vectors. Do you think you would mind adding another sentence to the explanation to clarify this difference?

orchid00:
OK, what about this:
If we use paste to concatenate two vectors (e.g. **a** and **b**), by default each item from vector **a** will be concatenated by a space, followed by a corresponding item from vector **b**, forming a new vector of paired values. That is using paste(a, b), which is exactly the same as using paste(a,b,sep=" "). If we want to add a different separator than space, like a comma we use sep=",". On the other hand, if we want to concatenate all items from both vectors into one character, then we use the attribute collapse, try paste(a,b, collapse=" "). The items will be concatenated one by one in pairs, and form a whole new character. We can also use both attributes together, that is sep and collapse, then we will replace the default space in between items for another separator like a comma, and collapse by using a hyphen. That will be paste(a, b, sep=",", collapse="-"), you can then try using different separators and collapsing characters. Remember that when using collapse you will get one whole character instead of a vector or characters.

naupaka:
That seems a little long - I was thinking more along the lines of something line:

`paste()` can take two arguments, `sep` and `collapse`. `sep` is for setting the string that goes between the components of each pair of elements, while `collapse` is for setting the string that goes between the elements in the set of joined pairs. If you are pasting together two vectors of length one, then the arguments produce the same output.

Example:
```r
> paste(c('a', 'b', 'c'), c('x', 'y', 'z'), sep = '%', collapse = ",")
[1] "a%x,b%y,c%z"

> paste(c('a', 'b', 'c'), c('x', 'y', 'z'), sep = ',', collapse = "%")
[1] "a,x%b,y%c,z"

> paste(c('a', 'b', 'c'), sep = "%")
[1] "a" "b" "c"

> paste(c('a', 'b', 'c'), collapse = "%")
[1] "a%b%c"

> paste('a', 'b', sep = " ")
[1] "a b"

> paste('a', 'b', collapse = " ")
[1] "a b"
```

naupaka:
hi @orchid00 - do you think you would be willing to add in the additional explanation as part of this PR?

orchid00:
Hi @naupaka If this can wait till the weekend I'll do that this Saturday. I have a busy week now.

naupaka:
Sure, that would be fine. No worries.

orchid00:
Hi @naupaka  I made some updates, Can you merge with this? https://github.com/orchid00/r-novice-gapminder/pull/1 
I didn't make the pull request on the original because it only gave the option of commit. Let me know if this is alright.

PR #107, opened 2016-05-10
Fix footer 'Source' link

lexnederbragt:
Change footer link 'Source' from the workshop template to the lesson repo at https://github.com/swcarpentry/r-novice-gapminder


naupaka:
Good catch here too. Thanks for the PR @lexnederbragt!

PR #109, opened 2016-05-12
change 'double' to 'integer' in data structures lesson

Myfanwy:
Correct output after appending the feline-data.csv with the line suggested in the lesson.

### DIFF
--- a/04-data-structures-part1.md
+++ b/04-data-structures-part1.md
@@ -179,7 +179,10 @@ cats$weight[1] + cats$weight[2]
 
 
 ~~~{.output}
-[1] 7.1
+[1] NA
+Warning message:
+In Ops.factor(cats$weight[1], cats$weight[2]) :
+  ‘+’ not meaningful for factors
 
 ~~~
 


naupaka:
Thanks @Myfanwy for the PR. This is the same problem @rgaiacs filed in issue #111. Since the output in these files is generated from the R code, simply changing the output to match what it should be will get overwritten the next time the lessons are knitted. To fix this we would need to add a second data file named (for example) `feline-data-2.csv` to the repo and then read that in instead to show the problem with text in a numeric column. So, I am not planning to merge this PR as is, but if you would like to fix the problem by adding the second data file and adjusting the lesson accordingly, I would be happy to merge that in (and it would resolve #111 as well).

naupaka:
This issue has been resolved by #114 so I'm closing this. Thanks for the effort, though!

Myfanwy:
No problem!  sorry,  should have read all the issues first.

On Mon, May 16, 2016 at 9:25 PM Naupaka Zimmerman <notifications@github.com>
wrote:

> This issue has been resolved by #114
> <https://github.com/swcarpentry/r-novice-gapminder/pull/114> so I'm
> closing this. Thanks for the effort, though!
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/swcarpentry/r-novice-gapminder/pull/109#issuecomment-219616182>
>
-- 
Myfanwy Johnston
Ph.D Candidate, UC Davis
Animal Behavior Graduate Group
Biotelemetry Laboratory

Issue #111, opened 2016-05-14
Please not change the CSV file on the fly for the lesson


rgaiacs:
Screenshot of `04-data-structures-part1.html`

![screen shot 2016-05-14 at 12 21 24](https://cloud.githubusercontent.com/assets/1506457/15268183/71b71fa6-19ce-11e6-9fdd-b2fe40f62a0d.png)

# Description

The lesson has

> Go back to your text editor and add add this line to feline-data.csv:
>
> ~~~
> tabby,2.3 or 2.4,TRUE
> ~~~
>
> Reload your cats data like before, and check what type of data we find in the weight column:
> 
> ~~~
> cats <- read.csv(file="data/feline-data.csv")
> typeof(cats$weight[1])
> ~~~
>
> ~~~
> [1] "double"
> ~~~
>
> Oh no, our weights aren’t the double type anymore! If we try to do the same math we did on them before, we run into trouble:
>
> ~~~
> cats$weight[1] + cats$weight[2]
> ~~~
>
> ~~~
> [1] 7.1
> ~~~

The text doesn't match with the code examples. We should use `data/feline-data2.csv` or something like that to avoid the problem.

naupaka:
Fixed by #114

Issue #110, opened 2016-05-14
Duplicate heading on 04-data-structures-part1.Rmd


rgaiacs:
~~~
$ git log --oneline -1
0c89e17 Updating HTML
$ grep -n "## Factors" 04-data-structures-part1.Rmd
226:## Factors
241:## Factors
~~~

Having this headings with the same name doesn't make sense to me. I think that the first one should be "Data Frame".

naupaka:
Closed by #114

PR #114, opened 2016-05-16
Modified data file for 04-data-structures-part1

tomwright01:
Addressed issue #111 - Asking users to manually modify a data file caused inconsistent output in the html.
Also issue #110 (duplicate headers) while I was in the file.


naupaka:
Excellent. Thanks!

Issue #100, opened 2016-04-07
08-plot-ggplot2 Switch axis for life exp. & GDP?


jcoliver:
The examples show GDP on the y-axis and life expectancy on the x-axis.  As many discussions (elsewhere) concern the utility of GDP in _predicting_ life expectancy, should GDP instead be on the x-axis?

jcoliver:
BTW, what do I win for opening the 100th issue?

naupaka:
I agree. Submit PR with fix? And for 100th issue: 👍  😄

naupaka:
Closed by #117

PR #117, opened 2016-05-24
Switch x and y axis so lifeExp on y, gdpPercap on x

jcoliver:
Per issue #100 , I switched the axes of some of the graphs so GDP per capita is a _predictor_ for life expectancy.  The knit'd .html was not committed (per .gitignore & point 3 of guidelines), so that would need to be added.

### DIFF
--- a/08-plot-ggplot2.Rmd
+++ b/08-plot-ggplot2.Rmd
@@ -52,7 +52,7 @@ Let's start off with an example:
 
 ```{r lifeExp-vs-gdpPercap-scatter, message=FALSE}
 library("ggplot2")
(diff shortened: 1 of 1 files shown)


naupaka:
HI @jcoliver - thanks for this! We usually only merge in changes to the Rmd. Do you think you could update the PR by removing the rendered png files? Those will get generated when I rebuild the lessons and it keeps the logs neater.

jcoliver:
Will do.  I'll let you know when I have cleaned up the PR.

jcoliver:
OK, PR #117 only includes the updated .Rmd file.  Let me know if I need to do more.

naupaka:
Looks good, thanks!

PR #119, opened 2016-05-25
Yl edit - SWC/DC training check out

ChemLibrarian:
This pull request is for fulfilling the requirement of the SWC/DC instructor certification. My name is Ye Li and my email is liye@umich.edu . I attended the training workshop between April 13-14 at Ann Arbor, Michigan. 

I added a figure to show a simple project layout and explain each item. Thank you for reviewing my pull request.


naupaka:
Hi @ChemLibrarian - thanks for the PR. I am a little concerned that the level of content you describe in this directory is quite advanced for where learners are likely to be at the beginning of the R material (e.g. they are likely not at all familiar with markdown, much less R markdown at this point). I like the idea of a visual project directory, but my sense is that it would likely be better to introduce a simpler version so early in the lessons - maybe focusing on a `data` directory, a `code` directory, a README, plus the various files that RStudio generates in association with a project?. Thoughts @aammd @tomwright01?

ChemLibrarian:
Thank you, Naupaka. I was thinking the same about simplicity when I was
making the screenshot. That's the version I came up with as minimal but
still deliver the necessary structure for reproducible analysis from my
perspective. I am open to suggestions to further simplify it.

Best,

Ye

naupaka:
I was thinking perhaps focus on an example with just `.R` files instead of having the `.Rmd` and the generated `.html` files. You could add an additional `src` or `code` folder to hold R scripts. I think I would also remove the `.Rdata` file, since we generally encourage beginners not to save their environments. Also technically the `.Rhistory` and `.Rdata` files are unrelated to version control, although I imagine you were trying to keep the explanations simple by grouping those together.

This may all sound a little nit picky, but in my experience teaching novices, it is always better to err on the side of less content, explained more fully, than more content only briefly mentioned.

tomwright01:
I really like this idea. However I agree with @naupaka about the need for simplicity at this stage.
I think the key message to get across is the seperation of data files, analysis scripts and outputs. That is a +1 to inclusion of a `src` folder as previously suggested. 
I'm not sure about the inclusion of .git/ and .RHistory as these will (should) exist in any 'real-world' environment. How about labeling them "System files required for project management"?

Also, you could use a slightly clearer, and more generic path for the folder i.e. `/home/user_name/projects/current_project/` rather than the current `SWC/SWCCertification/...` path.

aammd:
I like the idea of including a screenshot. I also agree with @naupaka and @tomwright01 that a simpler structure is better here. 
I'd even prefer something more colloquial than `src`, like `analysis` or `Rscripts`.

I think a screenshot similar to the following example from the [rrrpkg](https://github.com/ropensci/rrrpkg) repository would be helpful:

```
project
|- DESCRIPTION          # project metadata and dependencies 
|- README.md            # top-level description of content and guide to users
|
|- data/                # raw data, not changed once created
|  +- my_data.csv       # data files in open formats such as TXT, CSV, TSV, etc.
|
|- analysis/            # any programmatic code 
|  +- my_scripts.R      # R code used to analyse and visualise data 
```

ChemLibrarian:
Many thanks for the suggestions, All. I will make the modifications when I
am back to office from a conference.

Best,

Ye

naupaka:
@ChemLibrarian any further modifications on this? The PR has been open for a while now, and the lessons have changed quite a bit since your original PR. Let me know if you have some work in progress... If not, it might be easier to close this PR and have you open up a new one against a current copy of the lesson materials.

ChemLibrarian:
Sorry about the delay. I have been busy with a job change. Yes, please
close this one and I will open up a new one when I get a chance.Thank you
very much!

Best,

Ye

On Sat, Jul 30, 2016 at 3:08 PM, Naupaka Zimmerman <notifications@github.com

naupaka:
Thanks and we look forward to your revised PR once you have the time to get to it!

PR #122, opened 2016-05-27
Fix on example output of 04-data-structures-part1.md

diraol:
The modified part was copied from the previous example, but it needed to be changed with the new output.

### DIFF
--- a/04-data-structures-part1.md
+++ b/04-data-structures-part1.md
@@ -165,7 +165,7 @@ typeof(cats$weight[1])
 
 
 ~~~{.output}
-[1] "double"
+[1] "integer"
 
 ~~~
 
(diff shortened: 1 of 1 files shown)


naupaka:
Thanks for pointing that out @diraol. I think that the incorrect output will be fixed when I regenerate the output and lessons from the `.Rmd` files, which only happens every few weeks (instead of after every merged PR). I *think* that should fix it so I'm going to hold off on merging this because that would overwrite these changes anyway. If after regenerating the output, the error is still there, then I'll merge the fix.

diraol:
Hi Naupaka, just a question.
Should I edit the .md or the .Rmd file to send pull requests?

Thanks!

Diego Rabatone Oliveira
Em 28 de mai de 2016 10:23 AM, "Naupaka Zimmerman" <notifications@github.com>
escreveu:

naupaka:
@diraol please just edit the `.Rmd` files for all proposed changes. The `.md` files are generated from the `.Rmd` when I update/knit all the lessons locally, so any changes in them will get overwritten. Thanks for asking!

tomwright01:
This issue is solved by regenerating the .md and .html files. I'm going to close it. Thanks @diraol for spotting the issue.

Issue #118, opened 2016-05-24
08-plot-ggplot2 Subset data before creating multi-panel figure?


jcoliver:
Subsetting the data to countries with names starting with the letter "A" might provide a more visually appealing demonstration of the multi-panel figure, while preserving the utility of by-continent coloring (all five continents/colors are represented in the set of seven "A"-countries).

naupaka:
I agree, good suggestion.

naupaka:
Closed via #124

PR #124, opened 2016-06-02
Subset data for multi-panel figures

jcoliver:
Multi-panel figures were just a bit too cluttered when using the entire gapminder data set.  This change uses a subset of those data (countries starting with letter A or Z).  The goals of the subset were to (1) make the figure a bit more readable, (2) ensure all continents were represented (Oceania is a pain), (3) keep it nice by having a full matrix of graphs ("A" alone creates 7 graphs, resulting in two rows of three, and one row of one, less than pretty).  An alternative for a single letter would be "C" (4 x 4), but only four continents are represented (@#$* Oceania).

### DIFF
--- a/08-plot-ggplot2.Rmd
+++ b/08-plot-ggplot2.Rmd
@@ -211,10 +211,12 @@ variables and their visual representation.
 
 Earlier we visualised the change in life expectancy over time across all
 countries in one plot. Alternatively, we can split this out over multiple panels
-by adding a layer of **facet** panels:
+by adding a layer of **facet** panels.  Focusing only on those countries with 
+names that start with the letter "A" or "Z":
 
 ```{r facet}
-ggplot(data = gapminder, aes(x = year, y = lifeExp, color=continent)) +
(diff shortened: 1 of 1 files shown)


jcoliver:
Oh, this PR is in response to issue #118.

naupaka:
Thanks for this PR @jcoliver. Do you think we need to introduce the `%in%` and `substr()` syntax before using it here? I don't think (but I could be wrong) it is mentioned elsewhere in an earlier lesson. Or perhaps add a quick sentence to explain what those are doing (I'm thinking particularly with the `substr()` approach)?

jcoliver:
Great points.  `%in%` shows up in lesson 06 (subsetting data, Subsetting by name section), but I think you're right about `substr()`.  Let me do some more research and update the PR as appropriate.

jcoliver:
I added some language explaining `substr` and `%in%` as a "Tip".  I am not sure that is the best way to add the explanation, but I did not want it to be too distracting.  If it make more sense to add the explanation in the regular text flow, let me know and I will update the PR.

naupaka:
Thanks for making that change. Merging...

PR #125, opened 2016-06-06
adding-advanced-challenge-ggplot2

noracmitchell:
Adding advanced challenge to ggplot2 section, building on challenge 4 to change point shape and color by continent.

### DIFF
--- a/08-plot-ggplot2.Rmd
+++ b/08-plot-ggplot2.Rmd
@@ -207,6 +207,14 @@ variables and their visual representation.
 > Hint: do not use the `aes` function.
 >
 
+> ## Advanced Challenge {.challenge}
+>
+> Modify Challenge 4 so that the points are now a different shape and are 
+> colored by continent with new trendlines.
+>
+> Hint: This code will differ from Challenge 4.
+>
+
 ## Multi-panel figures
 
 Earlier we visualised the change in life expectancy over time across all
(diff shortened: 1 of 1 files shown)


naupaka:
Thanks for the PR. I'm not sure I've seen advanced challenges elsewhere -- but I don't see why they couldn't be included. Perhaps to keep things better organized you could change the prompt for the challenge and the solution to this challenge to both be text contained within the challenge 4 block and the challenge 4 solution block instead of as separate chunks. Otherwise I think the organization could start getting rather confusing as to which challenge is related to what.

noracmitchell:
Thanks for the feedback, I'm doing this as part of the instructor trainee checkout process. There's an advanced addition to Challenge 5 below, however this one just required an additional bit of code, whereas the advanced one I propose requires taking pieces out from the normal Challenge 4 solution. I was basing this one off of the advanced challenge in the dplyr topic in this same lesson.

naupaka:
Ah, I see. I guess my confusion is that in that case (with dplyr) the advanced challenge is the last in the lesson and goes at the very end, whereas this one is an add-on to a challenge in the middle of the lesson. In that case the advanced challenge is also an attempt to synthetically bring together a number of the things taught throughout the lesson. I think if you'd like to add an advanced challenge perhaps move it to the end and make it synthetic, or else add in a stand-alone challenge in the middle and update the numbers of the challenges that follow accordingly.

noracmitchell:
Great point, I thought it might be helpful to have a mid-way challenge for more advanced students who are breezing through the lesson, but a synthetic advanced challenge at the end might make more sense.

gvwilson:
I think this is fine - all we have to do is take out the word 'Advanced', but we can do that after merging.

tomwright01:
I have merged this manually. For some reason github hasn't noticed.

PR #126, opened 2016-06-06
Add alpha to scatterplot

tem11010:
Adding an alpha value to scatterplot, to introduce transparency. Useful for large datasets.

### DIFF
--- a/08-plot-ggplot2.Rmd
+++ b/08-plot-ggplot2.Rmd
@@ -99,7 +99,6 @@ ggplot(data = gapminder, aes(x = gdpPercap, y = lifeExp)) +
 > Hint: the gapminder dataset has a column called "year", which should appear
 > on the x-axis.
 >
-
 > ## Challenge 2 {.challenge}
 >
 > In the previous examples and challenge we've used the `aes` function to tell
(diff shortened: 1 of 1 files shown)


naupaka:
Thanks for the PR. 

One problem with adding in the `alpha` parameter here is that all of the examples further down in the lesson build on the same code as it is currently. So if you add an alpha parameter here and then don't also include it in every later exercise that builds on it in the series, then it is quite jarring to be flipping back and forth instead of building gradually in a stepwise fashion. I also think it's better to add only one new thing at a time as people are learning the syntax (so first adding `scale_x_log10()` and then adding `alpha = 0.5` in an additional example).

So if you'd like to have the alpha example included (which is nice to show people, I agree), then I would ask that you break it into a separate step of the figure elaboration series (vs including it as a second change in the scale step), and also change all of the code for the figures which follow it, for consistency.

tem11010:
Hi Naupaka

Thanks for the response.

I think you're right that giving people too many new things at once is confusing. I guess I was just hoping to prime people for seeing the alpha function. Perhaps it would be better to just add an explanation of the alpha function to challenge 5 instead? I feel like it is a really useful function to know.

Cheers, 
Tim

naupaka:
Yes I like that idea. It's already there and fits better in the overall flow I think. Let me know if you'd like me to close this PR and you can open another one with that change, or if you'd prefer to modify this one instead.

tem11010:
I think close this PR and I can open another one if you'd like.

naupaka:
Will do, thanks for the quick responses.

Issue #94, opened 2016-03-04
tidyr lesson - inconsistencies in dataframe atomic types


thesarahmeister:
When you load the gap_wide data from the .csv file in the repo, columns 37 and 38 get inputted as integers and columns 1 and 2 get inputted as factors. When you follow along in the lesson the gap_wide_new columns 1 and 2 are characters and columns 37 and 38 are numeric. This means the all.equal() function wont work until you make those changes in the gap_wide dataframe.

tomwright01:
I think this is now resolved by #136. If not please let me know and we will open this issue again.
Thanks everyone.

PR #136, opened 2016-06-13
Load the gap_wide dataset for tidyr lesson

mickley:
Added R code to load the gap_wide dataset in Data Manipulation with tidyr.
This dataset is referenced in the lesson without being loaded.
Loading the data with read.csv() causes problems later (see #94) because
columns are loaded as Factor not chr.

Loading the dataset with stringsAsFactors = FALSE (referenced in comments on line 26 of the Rmd)
fixes the problem and the all.equal() function works as intended.
Columns 37 and 38 are still imported as int instead of num, but this does not lead to problems.


tomwright01:
Good spot, thanks.

Issue #120, opened 2016-05-25
`if` syntax is used in lesson 07-functions before lesson 10-control-flow


stevenhwu:
Based on the current order, lesson `07-functions` will be covered before `10-control-flow`.
However, in the second half of the lesson, the `if` statement is used in the function, which is not properly introduced until lesson `10-control-flow`
```
calcGDP <- function(dat, year=NULL, country=NULL) {
  if(!is.null(year)) {
    dat <- dat[dat$year %in% year, ]
  }
  ...
```
It might make more sense to switch the order around. There are several ways to get around this, here are just a few of them;
- Design a different example without `if`
- Switch lesson 7 and 10
- Introduce `if`, follow by functions, then `for`
- Revisit functions after `if` and `for`.

naupaka:
I generally teach control flow before functions, so perhaps switching 7 and 10 would work well. Someone would have to go through them carefully to make sure there aren't any other content dependencies that are broken by changing the order.

marcsze:
Umich is taking a shot at this issue

marcsze:
There is a conflict with 1issue #129 but the extra recommended changes made in issue #137 should help resolve some/most of these

tomwright01:
I think this is now resolved. Thanks @marcsze

PR #137, opened 2016-06-13
Issue 120 - Swap 07-functions / 10-control-flow

marcsze:
We created two new .Rmd files where the old lesson 7 (functions) is now lesson 10 and the old lesson 10 (control flow) is now lesson 7.


tomwright01:
This causes a small problem with 09-vectorisation. The solution to challenge 4 in that file suggests using functions. I think the extra parts to the solution could be removed.

Also index.md and reference.md will need updating. That's going to cause conflicts with PR #129.

marcsze:
@tomwright01 for issue #120 I updated the requested files and will look at the problem with 09-vectorisation.

marcsze:
@tomwright01 made the change for the small problem with challenge 4 in lesson 9 now as well. Most are in #120 but this last one is part of #137 as well.

tomwright01:
This looks great, I'm going offline for an hour or so. If no-one else comments I'll merge when I get back.

marcsze:
@tomwright01 sounds good to me. Let me know if you need anything else.

PR #140, opened 2016-06-13
Fix minor typos and formatting

jrnold:
- fix misspellings
- format all R package names so they are bold
- format all R objects and functions to be typewriter font
- "dataframe" to "data frame"

PR #142, opened 2016-06-13
Rm subset from data aammd

aammd:
This closes #95 . Includes Andy's changes and makes a few minor editorial tweaks


ateucher:
Cool, nice work @aammd! Looks good to me - in fact it doesn't look like there are any open PRs that this should conflict with (Except perhaps #140 and #144)...

aammd:
Awesome, thanks for checking @ateucher (and thanks @brunogrande for checking these files) !! 😹

PR #144, opened 2016-06-13
Edit 04-data-structures-part1.Rmd

jrnold:
Two minor edits to 04-data-structures-part1

- add tip that you can edit/create text files (in this case a csv) within RStudio
- note that data frames are lists in which each vector has the same length

### DIFF
--- a/04-data-structures-part1.Rmd
+++ b/04-data-structures-part1.Rmd
@@ -12,7 +12,7 @@ source("tools/chunk-options.R")
 > ## Learning Objectives {.objectives}
 >
 > - To be aware of the different types of data
-> - To begin exploring the data.frame, and understand how it's related to vectors, factors and lists
+> - To begin exploring the data frame, and understand how it's related to vectors, factors and lists
 > - To be able to ask questions from R about the type, class, and structure of an object.
 >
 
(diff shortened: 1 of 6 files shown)


aammd:
Hi @jrnold ! Thanks for your comments so far! 
Unfortunately we were editing `04-data-structures-part1.Rmd` just as you were submitting this, so there is now some new material for you to merge into your branch before submitting your PR. Sorry about the timing! 😅 

Additionally, I think for the time being we'll stick with calling them `data.frame`s throughout the text. I think the reasoning behind this is to match the output of `class()`. It may also help to get learners to focus on a single new term ("data.frame") for thinking about their data. (though perhaps others want to chime in here?)

jrnold:
No problem, I can always rebase as needed. 

I'll see what others say about the `data.frame` v. "data frame", but it seems if data.frame is used to formally reference the object class, then it should be `data.frame`. Though my preference would be to follow the R conventions.

(Sorry about all the commits. I am giving the instructor 5-min presentation on this one tomorrow, so I figured I'd kill two birds with one stone and read over the materials carefully)

aammd:
@jrnold Sounds good. I think you will need to rebase here, or otherwise close this PR and open one with just the changes to `04-data-structures-part1.Rmd`, leaving the discussion of data frame to #145 .
There are also some updates to `04-data-structures-part1.Rmd` that you should include first, please 😅

jrnold:
Got rid the of the 'data.frame' to 'data frame' stuff. I hadn't meant that to be in this PR anyways.

tomwright01:
I really like the note on data frames. It clarifies nicely.

@jrnold The data frame commits are still in this PR. i.e. you commit it, then revert it. Would you mind submitting another PR with only the new tip and note about data frames.

PR #145, opened 2016-06-13
Data.frame to data frame

jrnold:
- Replace instances of data.frame with "data frame" which is how the R documentation refers to them; see `?data.frame`.
- Note that data frames are lists in which each vector has the same length
- Add tip that csv files can be created/edited within RStudio as a new text file.

### DIFF
--- a/04-data-structures-part1.Rmd
+++ b/04-data-structures-part1.Rmd
@@ -12,7 +12,7 @@ source("tools/chunk-options.R")
 > ## Learning Objectives {.objectives}
 >
 > - To be aware of the different types of data
-> - To begin exploring the data.frame, and understand how it's related to vectors, factors and lists
+> - To begin exploring the data frame, and understand how it's related to vectors, factors and lists
 > - To be able to ask questions from R about the type, class, and structure of an object.
 >
 
@@ -110,7 +110,7 @@ Oh no, our weights aren't the double type anymore! If we try to do the same math
 cats$weight[1] + cats$weight[2]
 ```
 
-What happened? When R reads a csv into one of these tables, it insists that everything in a column be the same basic type; if it can't understand *everything* in the column as a double, then *nobody* in the column gets to be a double. The table that R loaded our cats data into is something called a *data.frame*, and it is our first example of something called a *data structure* - things that R knows how to build out of the basic data types. In order to successfully use our data in R, we need to understand what these basic data structures are, and how they behave. For now, let's return to our original cats data and reload it, while we investigate this behavior further:
+What happened? When R reads a csv into one of these tables, it insists that everything in a column be the same basic type; if it can't understand *everything* in the column as a double, then *nobody* in the column gets to be a double. The table that R loaded our cats data into is something called a *data frame*, and it is our first example of something called a *data structure* - things that R knows how to build out of the basic data types. In order to successfully use our data in R, we need to understand what these basic data structures are, and how they behave. For now, let's return to our original cats data and reload it, while we investigate this behavior further:
 
 feline-data.csv:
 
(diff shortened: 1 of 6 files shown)


aammd:
I had originally commented on this over in #144 , but this is a better place for this whole discussion! my original comment below (supporting the current way of doing things, i.e. writing `data.frame`)

> I think the reasoning behind this is to match the output of `class()`. It may also help to get learners to focus on a single new term ("data.frame") for thinking about their data. (though perhaps others want to chime in here?)

jrnold:
And my comment from #144 

> I'll see what others say about the data.frame v. "data frame", but it seems if data.frame is used to formally reference the object class, then it should be data.frame. Though my preference would be to follow the R conventions.

gvdr:
On my side, as a user I'm +1 for "data.frame". Almost everywhere when we use the term, we are referring to objects of the specific class data.frame. See for example the use in dplyr, where it is particularly sensitive as there is a `data_frame()` function: https://cran.r-project.org/web/packages/dplyr/vignettes/data_frames.html

tomwright01:
The example actually uses data.frame() to refer to the function. When referring to the object type it uses 'data frame'.

gvdr:
Gosh, you are right. The dot was just in my brain. Much cognitive bias. Pardon me. Then, I'll second you.

aammd:
It looks like `data frame` might be winning, both on here and in my informal polls (real life and on Twitter)

gavinsimpson:
The concept of a data frame extends beyond R's `"data.frame"` class. See the `"tbl_df"` class in R package *tibble* and the `"DataFrame"` class in Python library Pandas. Julia also has a `"DataFrames"` package. It will be important going forward to talk about data frames in the abstract sense that cuts across implementations and even languages.

Hence I'm for "data frame" when discussing the concept, or talking about the spreadsheet-like tabular data objects. When referencing specific class names (which one rarely needs to do) only then should we resort to `data.frame`, or `DataFrame`.

aammd:
hi @gavinsimpson , thanks for your input! Out of curiosity, is there a language-agnostic description of what is meant by  a "data frame" (in the broad sense) that we could link to?

gvdr:
Hadley's paper on Tidy Data is quite language-agnostic, at least for section 2. *Defining tidy data* http://vita.had.co.nz/papers/tidy-data.pdf

jrnold:
Feather the Wes McKinney and Hadley Wickham uses the term 'data frames" https://blog.rstudio.org/2016/03/29/feather/ and is explicitly language agnostic. I can't find an explicit definition of data frame though.

jrnold:
Sorry, [here's](https://blog.rstudio.org/2016/03/29/feather/) a definition

> data frames are lists of named, equal-length columns, which can be numeric, boolean, and date-and-time, categorical (factors), or string. Every column can have missing values.

aammd:
@jrnold I like that! What if we add that definition to the lesson material as part of this PR?

tomwright01:
Please add the definition to the glossary in reference.MD

gavinsimpson:
That definition is not broad enough for the usage that Hadley and co are now espousing. Effectively, because data frames are lists, anything can be a component (as long as each component has the correct length), and so they are using nested data frames. From an EDA viewpoint, the examples hadley gave in his recent Edinburgh talk of nested data frames were quite persuasive.

That perhaps goes more towards the R implementation and perhaps `tbl_df`s and `the tidy data paradigm in particular.

jrnold:
Would replacing "numeric, boolean, and date-and-time, categorical (factors), or string" with "heterogeneous data types" handle that generalization?

gavinsimpson:
@jrnold yes, I guess so - even without the nested data frame distraction, R has more data types than the description from the RStudio blog (e.g. raw, complex, if we're being extra picky), which may not be the same in other implementations.

jrnold:
This is the response I got from twitter https://twitter.com/thosjleeper/status/743178779212787713

> A data frame is a tabular data structure. Usually, it contains data where rows are observations and  columns are variables of various types. While "data frame" or "dataframe" is the term used for this concept in several languages (R, Apache Spark, deedle, the pandas library in Python and the DataFrames library in Julia), "table" is the term used in MATLAB and SQL.

The source is [stackoverflow](http://stackoverflow.com/tags/dataframe/info) tag info for `dataframe`.

jcoliver:
Closing for now given upcoming discussion around tidyverse.

PR #146, opened 2016-06-13
Remove just

jrnold:
Purge instances of the word just from the lessons


tomwright01:
I think this is a good set of changes that improves the readability of the lessons.
Could you _just_ rebase this onto the current materials?

jrnold:
rebased

PR #149, opened 2016-06-13
Add line for appropriate Tip formatting

jcoliver:
The second Tip in the lesson (any and all) was not being formatted correctly.  I added an empty line between the tip header and content in an effort to get the formatting to work (thumb-tack image, blue box, etc.).

### DIFF
--- a/07-control-flow.Rmd
+++ b/07-control-flow.Rmd
@@ -119,7 +119,8 @@ If your condition evaluates to a vector with more than one logical element,
 the function `if()` will still run, but will only evaluate the condition in the first
 element. Here you need to make sure your condition is of length 1.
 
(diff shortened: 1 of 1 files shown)


tomwright01:
@jcoliver Unfortunately inserting the empty line didn't solve the problem. I think it's the back-ticks in the title that are causing the problem. If there isn't a nicer way these could just be removed.

jcoliver:
@tomwright01 Got it.  I just removed the back-ticks.

jcoliver:
Ah.  Hold off on the PR.  Was still using a branch where control flow was lesson 10...

naupaka:
Looks good, thanks.

Issue #139, opened 2016-06-13
04-data-structures-part1 Data.frames as mixed lists?


gvdr:
I find line 313 a bit confusing and potentially wrong.

> We see that data.frames look like lists 'under the hood' - this is because a data.frame is really a list of vectors and factors, as they have to be - in order to hold those columns that are a mix of vectors and factors, the data.frame needs something a bit more flexible than a vector to put all the columns together into a familiar table.

I think that the list in a data.frame are not of mixed type. Am I wrong here?

If not I'll propose a PR with a less problematic (imho) wording.

tomwright01:
With hindsight, I agree this wording could be confusing.

gvdr:
I've proposed a pull request for this. Not sure it's much better ;-)

aammd:
i do believe this was closed in #151 ?

PR #151, opened 2016-06-13
minor rewordings

gvdr:
Just some minor typos and some rephrasing to result more readable.

### DIFF
--- a/04-data-structures-part1.Rmd
+++ b/04-data-structures-part1.Rmd
@@ -310,7 +310,7 @@ We can now understand something a bit surprising in our data.frame; what happens
 typeof(cats)
 ```
 
-Under the hood, a data.frame is a list of variables, which can be of different kind. In our `cats` example, we have an integer, a double and a logical variable. As we habe seen already, ach column of data.frame is a vector.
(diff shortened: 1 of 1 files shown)


aammd:
Helpful! Thank you @gvdr 😄

aammd:
@gvdr Uh -oh it looks like you *do* have merge conflicts after all! I'll merge this PR once you bring in those changes

gvdr:
As for #139 I added a rewording of the bit about *under the hood* of data.frames.

04-data-structures-part1.Rmd
```diff
@@ -310,7 +310,34 @@ We can now understand something a bit surprising in our data.frame; what happens
 typeof(cats)
 ```
 
-We see that data.frames look like lists 'under the hood' - this is because a data.frame is really a list of vectors and factors, as they have to be - in order to hold those columns that are a mix of vectors and factors, the data.frame needs something a bit more flexible than a vector to put all the columns together into a familiar table.
+Under the hood, a data.frame is a list of variables, which can be of different kind. In our `cats` example, we have an integer, a double and a logical variable. As we habe seen already, ach column of data.frame is a vector.
```
tomwright01: Couple of types habe/have ach/each

gvdr:
Solved the ach habe typos.

PR #153, opened 2016-06-13
Update 07-functions.Rmd (updated)

scientificbruno:
I'm helping @aammd with some outstanding PRs. In this case, a pull request (https://github.com/swcarpentry/r-novice-gapminder/pull/101) was opened by @orchid00 and subsequent changes based on the PR discussion were made to a separate branch. This PR serves to consolidate all of @orchid00's changes as well as merge these changes with the latest commits on the repository. 

This PR uses the commits from https://github.com/orchid00/r-novice-gapminder/pull/1.

### DIFF
--- a/07-functions.Rmd
+++ b/07-functions.Rmd
@@ -290,19 +290,29 @@ which is much better than in our first attempt where we just got a vector of num
 >
 > ```{r}
 > best_practice <- c("Write", "programs", "for", "people", "not", "computers")
-> paste(best_practice, collapse=" ")
+> paste(best_practice, collapse = " ")
 > ```
+[1] "Write programs for people not computers"
 >
->  Write a function called `fence` that takes two vectors as arguments, called
+> *Note:* The function `paste` converts items into a character type and puts them together. It can be used for single elements, elements of one vector or multiple vector's elements. The function `paste` has three arguments:
+> `paste (..., sep = " ", collapse = NULL)`
+>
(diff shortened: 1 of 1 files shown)


aammd:
Hello Everyone! great discussion so far. @orchid00 I like your edits, and @naupaka I also like your [demonstration of the behaviours of `sep` and `collapse`](https://github.com/swcarpentry/r-novice-gapminder/pull/101#issuecomment-209117474)

However I wonder if both are not a bit too much for this section. Here we want to get learners to write their own function, and we just want them to know enough about how `paste` works to do so.

How about something like:

> `paste` is used for combining character vectors. Here are two examples of how it works:

```r
paste("a", "b", sep = "%")
```
vs

```r
paste(c("a", "b"), collapse = "%")
```

This way there is a lot less info about how to use `paste`, but just enough that the challenge is solvable.
 It would be nice to show both `collapse` and `sep` so that there is more than one "right" answer

orchid00:
I like that! if you want to keep it.
@aammd i also think that we deviated from the challenge getting too much info about paste. It is really unnecessary for beginners. Personally when teaching this lesson I only used sep, and I always tell the audience to check for help via ?functionname so they learn by themselves what other attributes might be worth using.

naupaka:
Also agree less is more here. Couple quick examples should be plenty.

naupaka:
I will work on resolving the merge conflicts on this one and merging it in after I get back from traveling in a few weeks.

PR #154, opened 2016-06-13
Reference Good Enough Practices for Scientific Computing

jrnold:
Add tip which references "Good Enough Practices for Scientific Computing" by Greg Wilson et al.

Not included in this PR, but perhaps the reference to ProjectTemplate should be removed?


aammd:
I like that. It is good to reference the paper and the discussion around it, rather than a specific tool to do that. See also the discussion in #119

tomwright01:
This is great and will be enhanced by the figure from @ChemLibrarian Issue #119.
Merging.

PR #156, opened 2016-06-13
Added solution Challeng 1,2

gvdr:
Added solutions for the first two challenges.

- [x] Challenge 1
- [x] Challenge 2
- [ ] Challenge 3
- [ ] Challenge 4
- [ ] Challenge 5

### DIFF
--- a/10-control-flow.Rmd
+++ b/10-control-flow.Rmd
@@ -15,8 +15,8 @@ set.seed(10)
 
 > ## Learning Objectives {.objectives}
 >
-> * Write conditional statements with `if` and `else`.
-> * Write and understand `for` loops.
+> * Write conditional statements with `if()` and `else()`.
+> * Write and understand `for()` loops.
 >
 
 Often when we're coding we want to control the flow of our actions. This can be done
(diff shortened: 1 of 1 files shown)


10-control-flow.Rmd
```diff
@@ -272,3 +272,67 @@ output_vector2
...
+> Putting all together, we obtain:
+> ```{r ch10pt4-sol, eval=FALSE}
+> if(nrow(gapminder[(gapminder$year == 2002),]) >= 1){
+>    print("Record found.")
+> }
+> ```
+
+> All this can be done more quickly with `any`. The logical condition can be expressed as:
+> ```{r ch10pt5-sol, eval=FALSE}
+> any(gapminder$year == 2002)
```
tomwright01: I like the way you go into extended detail with this solution.
Can you expand this part of the solution to include an if statement and a 'pretty' message?
```
if(any(gapminder$year) == 2002){
    print("Record found.")
}
```

10-control-flow.Rmd
```diff
@@ -272,3 +272,67 @@ output_vector2
...
+> ```
+
+> ## Solution to Challenge 2 {.challenge}
+> We can check whether the two vectors are identical using the `all` function:
```
tomwright01: When referring to functions in text I think our general approach is to add `()` to clarify the status as a function. I like this convention as it reinforces the difference between functions and variables.
Would you change `all` to `all()`. 
Also `as.vector()` in line 319 and `t()` in line 322.

Thanks, and great answers.

tomwright01:
Thanks for starting with these, very comprehensive answers. I've made a couple of comments inline before this can be merged.

gvdr:
Sorry, I couldn't read the comments until now. And I will not be able to do anything for a bit (until I reach my office). I do agree, it was just poor writing.

gvdr:
Just for clarification, would the standard work for conditionals and loops? Is it `if()`, `for()` or `if`,`for`?

tomwright01:
Not sure at the moment it's not really a standard, more a preference. @naupaka do you have any preference regarding the use of brackets for function names e.g. `for()` ?

naupaka:
I am a big fan of using the `function()` style for all things that are functions. I think we should encourage everyone to follow that convention when contributing. I just checked the [Google R style guide](https://google.github.io/styleguide/Rguide.xml) and they don't explicitly recommend it, as far as I can tell, but they do use that style themselves, so let's stick with it.

PR #160, opened 2016-06-14
Edit 04-data-structures-part1.Rmd

jrnold:
- Add note on editing text files (csv) within R
- Add note relating data frames to lists

This is like #144 without the find and replace of data.frame.


tomwright01:
Great, thanks.

PR #165, opened 2016-06-16
Reduced knitr lesson header levels from h3 to h2 

naupaka:
This is to maintain consistency with other lessons and to reduce warnings from `make check`.


naupaka:
Just a quick tweak. Look ok @tomwright01?

tomwright01:
Looks good to me, I actually noticed this while I was preparing the merge, however I forgot to do anything about it. Thanks for picking this up.

PR #182, opened 2016-09-14
R novice gapminder exercise amendment

KanduriC:
The first exercise in challenge-7 of 06-data-subsetting demonstrates a subsetting error, which assigns all the values of gapminder$year to 1957. When students do not get this exercise right, they will also get the final exercise in the same challenge-7 wrong (because there will be no values for gapminder$year == 2002 | gapminder$year == 2007). It would be nice to learn this combinatorial subsetting independent of the previous example. So I suggested an amendment.

### DIFF
--- a/_episodes/06-data-subsetting.md
+++ b/_episodes/06-data-subsetting.md
@@ -1489,8 +1489,8 @@ be changed with the third argument, `drop = FALSE`).
 > ~~~
 > {: .r}
 >
-> 5. Advanced: extract rows that contain information for the years 2002
->    and 2007
+> 5. Advanced: extract rows that contain information for the continents Americas
+>    and Asia
 >
 > 
 > ~~~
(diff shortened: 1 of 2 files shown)


naupaka:
Hi @KanduriC thanks for your suggestion. My sense of how the parts of challenge 7 would usually be taught is that it would be done without actually running the broken code, so I don't think there should be any issue due to the fact that the first bit of 'broken' code would overwrite the values in that column. 

Many instructors go through these challenges with the class as a group to explain the misconceptions that each represents and what to be careful of when doing such operations. I also think that it is important for students to be able to learn to troubleshoot when something goes wrong (e.g. they overwrite all values in a column). My inclination is to leave the lesson as-is for now.

naupaka:
I'll also make a quick note that the general flow of proposing changes to these lessons is to submit an issue (instead of a Pull Request) on this repository. Once the change is proposed in an issue, we can discuss the suggestion before any changes are made to the materials themselves. Then if people agree it's a good idea to make the change, someone can open up a pull request to address that issue.

KanduriC:
Thanks for the quick reply. I was a helper at a carpentry workshop yesterday, where more than half of the students got the first part wrong, which overwrote the column (they were given time to run the code). This resulted in no result for the last part irrespective of correct syntax. They realized what went wrong when the instructors and helpers ran through the solutions. I agree that it is totally important for the students to be able to learn to troubleshoot the errors, but the majority of audience we had were quite novices, who will have to arguably go through a bit of learning curve. I also agree that dependence between parts of the challenge could be beneficial in some cases.  Thanks.

naupaka:
Hmmm - thanks for the follow up. I see your point. What about adding in a short sentence at the top of the challenge to the effect of: "Before you run any of these lines of code (which are currently incorrect), think about what they would do. Remember, if you make an incorrect change to your dataset, you can always reload a fresh version of the data into R."?

naupaka:
thoughts @aammd @tomwright01 ?

naupaka:
I am going to close this for now @KanduriC, but if you'd like to add a sentence of warning to the lesson as described above then I'd be happy to reopen and merge it in.

PR #185, opened 2016-09-21
Update feline-data_v2.csv

adamhsparks:
This updates the feline-data_v2.csv to match feline-data.csv, replacing TRUE, FALSE with 0, 1 for the exercises.


naupaka:
Thanks!

Issue #184, opened 2016-09-20
feline-data.csv that is created does not match later example data in R session


adamhsparks:
In [04-data-structures-part1](http://swcarpentry.github.io/r-novice-gapminder/04-data-structures-part1/) the instructions say to create a new file using this data:
```
coat,weight,likes_string
calico,2.1,1
black,5.0,0
tabby,3.2,1
```

However, in the next lines, the data that is "imported" into R changes the `1, 0, 1` of `likes_string` to `TRUE, FALSE, TRUE`. R does not automatically do this when reading the CSV file, so I'm not sure where this comes from?

naupaka:
Good catch. That looks to me like a typo. It looks like it should be changes to `TRUE, FALSE, TRUE` instead of `1, 0, 1` in that chunk. @adamhsparks would you be willing to open a pull request to fix this issue?

ateucher:
I think they should actually be `1,0,1` right the way through, as they are used as an example of coercion from `numeric` to `logical` in the ` "Vecotrs and Type Coercion" Section:

> But coercion can also be very useful! For example, in our cats data likes_string is numeric, but we know that the 1s and 0s actually represent TRUE and FALSE (a common way of representing them). We should use the logical datatype here, which has two states: TRUE or FALSE, which is exactly what our data represents. We can 'coerce' this column to be logical by using the as.logical function:
>
> cats$likes_string
> cats$likes_string <- as.logical(cats$likes_string)
> cats$likes_string

naupaka:
Ah, yes, I think that's right @ateucher. So the proper fix would be to make them all 1s and 0s in that column in all the occurences in the code above line 300, where the `as.logical()` is used. Correct?

ateucher:
Yup, I think that's right @naupaka. The Rmd is a bit messy with numerous `cats` objects being created with `data.frame`, as well as a mix of both `read.csv` and `write.csv` calls, some with eval=FALSE and some being actually run. I can have a quick go at fixing it up if you like, unless @adamhsparks wants to do it?

naupaka:
@ateucher that would be excellent, thanks!

adamhsparks:
I've been working on it this morning. @ateucher  is right, it's a bit messy. I need to actually go through it and test it first to see if I've fixed everything properly before I'd submit a pull request.

naupaka:
Closed by #186

PR #186, opened 2016-09-21
Correct the "likes_string" column

adamhsparks:
The data are entered at 0, 1 but the text is shown as TRUE, FALSE. This doesn't happen until the example using `as.logical` on line 304 when it's coerced.


naupaka:
Thanks! closes #184

PR #199, opened 2016-09-27
Fix box

rgaiacs:
Before:

screenshot_2016-09-27_16-21-04

After:

screenshot_2016-09-27_16-22-09


rgaiacs:
Related with https://github.com/jekyll/jekyll/issues/4563.

Issue #173, opened 2016-07-20
bad link (gapminder)


cwmohr:
In section 5 data structures part 2 [http://swcarpentry.github.io/r-novice-gapminder/05-data-structures-part2/], the gapminder file hyperlink is down: https://swcarpentry.github.io/r-novice-gapminder/data/gapminder-FiveYearData.csv

The file is however accessible via https://raw.githubusercontent.com/resbaz/r-novice-gapminder-files/master/data/gapminder-FiveYearData.csv

However this needs to updated on the webpage.

lmsimp:
Hi @cwmohr, @gvwilson I have addressed in the above issue in this [pull request](https://github.com/swcarpentry/r-novice-gapminder/pull/204) as part of my SWC checkout part I and also issue #187 related to the above.

naupaka:
Closed by #204

Issue #187, opened 2016-09-22
Exploring Data Frames: Downloading gapminder files before reading them in?


RaoulWolf:
I noticed that after Challenge 1 in Lesson 5 (Exploring Data Frames), the gapminder dataset is supposed to be read in, either from the data subfolder, or directly from the URL. However, I would suggest downloading the data set first (not mentioned in the material at this moment!) and then go for option 1. In an existing RStudio project, the code could look something like this:

```
download.file("https://raw.githubusercontent.com/resbaz/r-novice-gapminder-files/master/data/gapminder-FiveYearData.csv", destfile = "data/gapminder-FiveYearData.csv")
gapminder <- read.csv("data/gapminder-FiveYearData.csv")
```

In my opinion this is good practice in order to have your data set available locally, and it also fixes the wrong gapminder URL.

naupaka:
@RaoulWolf yes I think we agree that would be a better approach. #172, #173, and this (#187) are all related to the same problem. I think a fix would include changing the path to the dataset to a URL for the gapminder website in episode 1, as suggested by @ashander, adding the `download.file()` line before the `read.csv()` line in episode 5 as suggested above and in #172. PRs welcome to fix these.

lmsimp:
Hi @naupaka, @RaoulWolf  I have attempted to address the above issue in this [pull request](https://github.com/swcarpentry/r-novice-gapminder/pull/204) as part of my SWC checkout part I and also issue #173 related to the above.

naupaka:
Closed by #204

Issue #172, opened 2016-07-20
Getting Gapminder dataset


cwmohr:
Currently there are two options presented for loading in the gapminder dataset in session 5 data structures part 2 [https://github.com/swcarpentry/r-novice-gapminder/blob/gh-pages/_episodes/05-data-structures-part2.md]:

The use of the `read.csv()` function:
`gapminder <- read.csv("data/gapminder-FiveYearData.csv")`,

in which case you have to have the file pre-downloaded to that folder. The file is available at https://swcarpentry.github.io/r-novice-gapminder/data/gapminder-FiveYearData.csv . However for MacOSX users, simply adding the link into the address bar results sometimes in the file opening in the browser (not sure why this happens at times). It is possible to save the file from the browser via the File menu clicking "save as", but this seems to be an unnecessary amount of hassel for the students.

Alternative two is a "direct read in":
`gapminder <- read.csv("https://swcarpentry.github.io/r-novice-gapminder/data/gapminder-FiveYearData.csv")`

This works perfectly (given that the link still exists). However, in other scenarios where the data file or files are massive. Re-running the R-script results in unnecessary re-downloading from site.

**Solution:**
One option that might be better, is to teach the students the `download.file()` function:
`download.file("https://swcarpentry.github.io/r-novice-gapminder/data/gapminder-FiveYearData.csv", "data/gapminder-FiveYearData.csv")` 

 By using this function the file can be downloaded to the project folder. Once there the file can be read in using the `read.csv()` function. By this method the student have learned both the `download.file()` and the `read.csv()` function.

cwmohr:
By the way the link to https://swcarpentry.github.io/r-novice-gapminder/data/gapminder-FiveYearData.csv is down. See #173

tomwright01:
I think the link should be:
https://swcarpentry.github.io/r-novice-gapminder/_episodes_rmd/data/gapminder-FiveYearData.csv

However, I can't make this work either.

I think I like the `download.file()` option, however we need to solve where to store data files.

gvwilson:
OK, I think we need @fmichonneau on this one:
1.  The source of the data is in `_episodes_rmd/data/*.csv` so that the R code in `_episodes_rmd` can read it using `data/whatever.csv` (the "natural" path).
2.  But `_episodes_rmd/*` isn't copied into `_site` when Jekyll builds the web version of the material.
[...]

PR #204, opened 2016-09-29
Fix issue187

lmsimp:
I have added a few new lines of code in this branch as suggested by issue #187 to address downloading the gap minder data, and fixed the URL in issue #173 in [section 5 data structures part 2](http://swcarpentry.github.io/r-novice-gapminder/05-data-structures-part2/). 

This is also to fulfil part I of the checkout procedure for the SWC instructors training (attended Cambridge 19/20th September).


lmsimp:
Yes, @ashander agreed no strong reason so have changed the URL as suggest in #172

ashander:
Great, it seems this issue has been discussed a number of different places. Hopefully the maintainers agree with our interpretation and it's to finally fix it :)

One thing: the second instance of the URL should also be changed: https://github.com/swcarpentry/r-novice-gapminder/pull/204/files#diff-2267c0f189e8cda7f265c0981ca5c273R269

After that it looks good to me

lmsimp:
@ashander sorry I missed that URL! Now changed in the above commit. In reference to issue #206 yes I believe the only .Rmd needs changing looking at the Makefile. I believe the maintainers regenerate everything following PR and changes.

naupaka:
Hi @lmsimp thanks for this! Could you also remove the `.md` file from the PR? They can cause issues with the makefile we use to generate the pages.

lmsimp:
Hi @naupaka I have removed my updated `.md` . Would you like me to add back in the original unchanged to this PR? Or will you regenerate anyway? Thanks.

naupaka:
No need to re-add, since the markdown gets generated when I rebuild the site. Thanks for the PR! Closes #173 and #187

snamburi3:
The episode 5 - "Exploring Dataframes" is missing from the main [site](http://swcarpentry.github.io/r-novice-gapminder/04-data-structures-part1/.) . The .md file does not exist in the episodes/ folder.

naupaka:
Thanks for catching that. Must have been a snag during the site rebuilding with the makefile. I'll work on redoing that ASAP.

PR #205, opened 2016-09-29
Link refererence to data in setup to gapminder.org

ashander:
Partial fix for the issue of referencing data before showing how to download (raised downthread in #172 )


naupaka:
Thanks for this fix!

PR #210, opened 2016-10-03
Add solution to exercise mentioning colClasses

rgaiacs:


naupaka:
Nice addition, thanks!

PR #212, opened 2016-10-03
Make key points more specifics

rgaiacs:
Because

~~~
> ?cbind  # Work
> ?rbind  # Work
> ?omit  # Fail with the following message
No documentation for ‘omit’ in specified packages and libraries:
you could try ‘??omit’
> ?na.omit  # Work
~~~


naupaka:
Good catch.

Issue #208, opened 2016-10-01
Improve diagrammer figures for episode 13 - dplyr


simonthelwall:
Currently the figures use Times New for the font, a monospace font might look better to reflect that the text is intended to be code. Also, lines between nodes in fig 2 and fig 3 would look better if smooth, as in fig 1.

naupaka:
The code that generates those figures is available [here](https://github.com/swcarpentry/r-novice-gapminder/blob/gh-pages/fig/13-dplyr-generate-figures.R) for the dplyr episode and [here](https://github.com/swcarpentry/r-novice-gapminder/blob/gh-pages/fig/14-tidyr-generate-figures.R) for the tidyr episode. A PR to change those scripts to use a monospaced typeface in those figures would be welcome.

simonthelwall:
I'd love to. and looked at the code last night before raising this as an issue. However, I've never used DiagrammeR before, nor graphviz and couldn't figure out how to alter the fontname of the label. If anyone can point me in the right direction I'll happily add the code.

naupaka:
I don't have any personal experience with it, but from the [DiagrammeR](http://rich-iannone.github.io/DiagrammeR/graph_creation.html) docs it looks like you can change things with the 'fontname' parameter.

simonthelwall:
Well I assumed that fontname would change it too and tried that without success. Also, these particular cases seem to use graphviz which has [labelfontname](http://www.graphviz.org/doc/info/attrs.html#d:labelfontname). I tried changing this too in the first figure again without success.

simonthelwall:
Note to self: maybe [this SO post](http://stackoverflow.com/a/6363817/2633645) will help.

naupaka:
Closed by #214

PR #214, opened 2016-10-04
Changed DiagrammeR label font

simonthelwall:
Changed DiagrammeR font from Times New to Courier for all three figures in episode 13 - dplyr. Also split label for fig 3 across three rows to make it more 'dplyr-like'. Fixes issue #208


naupaka:
Awesome, thanks! Would you mind making the same change to the figures in the tidyr episode, to keep things consistent?

simonthelwall:
Sure, ok.

PR #216, opened 2016-10-06
Add missing _episodes/05-data-structures-part2.md back

rgaiacs:
As reported at https://github.com/swcarpentry/r-novice-gapminder/pull/204:

> The episode 5 - "Exploring Dataframes" is missing from the main site . The .md file does not exist in the episodes/ folder.

This is true.

screenshot_2016-10-06_06-45-09

> Thanks for catching that. Must have been a snag during the site rebuilding with the makefile. I'll work on redoing that ASAP.

Makefile is fine at 4da7e6b. This is the version of _episodes/05-data-structures-part2.md that I build on my machine.

### DIFF
--- a/_episodes/05-data-structures-part2.md
+++ b/_episodes/05-data-structures-part2.md
@@ -1,0 +1,621 @@
+---
+title: "Exploring Data Frames"
+teaching: 20
+exercises: 10
+questions:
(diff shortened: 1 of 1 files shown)


naupaka:
Not needed because issue was resolved with #217

PR #225, opened 2016-10-29
ggplot2 lesson: add tip: using a value instead of a mapping for aesthetics

daattali:


naupaka:
I think this is a really important concept to mention, so thanks for the PR. Setting also occurs a few figures down in this lesson, with the `alpha=0.5` setting. I think it would be nice to add a sentence in the description of that plot as well, reinforcing the point you're making that anything outside of an `aes()` call is a setting instead of a mapping (many to one instead of many to many). Could you add such a sentence, @daattali? After that addition, I'm happy to merge.

daattali:
@naupaka good point! Done

naupaka:
Thanks!

PR #227, opened 2016-11-14
added set.seed() in lesson 07, code chunk line 68

r-lange:
set.seed() is mentioned in the description of what will be done prior to the code chunk and is then further explained in more detail after the code chunk. For learners to fully understand how the function is used it should be shown in action.


naupaka:
Thanks for catching this!

Issue #194, opened 2016-09-27
Use "combine function" instead of "concatenate function"


rgaiacs:
On 04-data-structures-part1 we find

> You can also make vectors with explicit contents with the concatenate function:

and I would like to use "combine function" instead of "concatenate function" 
since the official documentation use the word combine,

> This is a generic function which combines its arguments.

naupaka:
Sounds reasonable. A PR to fix would be welcome.

Issue #218, opened 2016-10-06
Learners can't use R 3.0.2


rgaiacs:
One of the learners on today workshop installed R 3.0.2 on his old Ubuntu. The learner couldn't install some R packages because they weren't available to that version of R. Can we state that on the lesson and https://github.com/swcarpentry/workshop-template/?

naupaka:
A PR stating that an up to date version of R is required would be welcome :)

PR #228, opened 2016-12-01
Fixes for #194 and #218

ksamuk:
Two quick fixes for issues #194 and #218. Not sure about the etiquette re: these being one PR or two separate PRs. Cheers!


naupaka:
Thanks! Closes #194 and #218

Issue #211, opened 2016-10-03
Should we try to avoid long lines?


rgaiacs:
On http://swcarpentry.github.io/r-novice-gapminder/05-data-structures-part2/, we have

```
df <- data.frame(id = c('a', 'b', 'c'), x = 1:3, y = c(TRUE, TRUE, FALSE), stringsAsFactors = FALSE)
```

Should we rewrite it as

```
df <- data.frame(id = c('a', 'b', 'c'),
    x = 1:3,
    y = c(TRUE, TRUE, FALSE),
    stringsAsFactors = FALSE)
```

to avoid long lines as well as make easy to learners see each one of the arguments?

naupaka:
Yes, I think that would always be preferable.

naupaka:
Fixed by #229

Issue #189, opened 2016-09-27
Combine function show up on 01-rstudio-intro


rgaiacs:
On lesson 01-rstudio-intro, the answer to the last question is

> ```
> install.packages(c("ggplot2", "plyr", "gapminder"))
> ```

but we didn't introduce the combine or `c``function.

I suggest that we replace the answer with

> ```
> install.packages(c("ggplot2"))
> install.packages(c("plyr"))
> install.packages(c("gapminder"))
> ```

naupaka:
Why not: 

```r
install.packages("ggplot2")
install.packages("plyr")
install.packages("gapminder")
```

If you include `c()` anyway it still a problem.

PR would be welcome to tweak this.

naupaka:
Fixed by #229

Issue #183, opened 2016-09-15
Rmd ordered list not rendering properly for episode 6


naupaka:
In [Episode 6 (Subsetting)](http://swcarpentry.github.io/r-novice-gapminder/06-data-subsetting/), the following code renders as a list of ones ("1. ") instead of incrementing. It looks like the markdown parser is seeing each numeral as the beginning of a new ordered list. Perhaps to fix the code chunks need to be indented within each list item?

``````
> ## Challenge 7
>
> Fix each of the following common data frame subsetting errors:
>
> 1. Extract observations collected for the year 1957
>
> ```{r, eval=FALSE}
> gapminder[gapminder$year = 1957,]
> ```
>
> 2. Extract all columns except 1 through to 4
>
> ```{r, eval=FALSE}
> gapminder[,-1:4]
> ```
>
> 3. Extract the rows where the life expectancy is longer the 80 years
>
> ```{r, eval=FALSE}
> gapminder[gapminder$lifeExp > 80]
> ```
>
> 4. Extract the first row, and the fourth and fifth columns
>   (`lifeExp` and `gdpPercap`).
>
> ```{r, eval=FALSE}
> gapminder[1, 4, 5]
> ```
>
> 5. Advanced: extract rows that contain information for the years 2002
>    and 2007
>
> ```{r, eval=FALSE}
> gapminder[gapminder$year == 2002 | 2007,]
> ```
>
``````

renders as:

> ## Challenge 7
> 
> Fix each of the following common data frame subsetting errors:
> 1. Extract observations collected for the year 1957
> 
> ``` r, eval=FALSE
> gapminder[gapminder$year = 1957,]
> ```
> 1. Extract all columns except 1 through to 4
> 
> ``` r, eval=FALSE
> gapminder[,-1:4]
> ```
> 1. Extract the rows where the life expectancy is longer the 80 years
> 
> ``` r, eval=FALSE
> gapminder[gapminder$lifeExp > 80]
> ```
> 1. Extract the first row, and the fourth and fifth columns
>    (`lifeExp` and `gdpPercap`).
> 
> ``` r, eval=FALSE
> gapminder[1, 4, 5]
> ```
> 1. Advanced: extract rows that contain information for the years 2002
>    and 2007
> 
> ``` r, eval=FALSE
> gapminder[gapminder$year == 2002 | 2007,]
> ```

naupaka:
Fixed by #229

PR #229, opened 2016-12-04
Fixes for issues #183 #189 and #211

HLindsay:
This pull request makes some minor formatting changes to fix: 
- Issue #183 - numbered list not rendering correctly in Challenge 7, lesson 7
- Issue #189 - Avoiding use of c() before it is introduced in lesson 1 Challenge 5
- Issue #211 - splitting data.frame commands to avoid long lines in lesson 5 Challenge 1 and solution

I'm a newcomer doing this as part of my instructor training.  I haven't run the make file to regenerate the md files.  Please let me know if I should also do this.

Helen

### DIFF
--- a/_episodes_rmd/01-rstudio-intro.Rmd
+++ b/_episodes_rmd/01-rstudio-intro.Rmd
@@ -573,9 +573,10 @@ network). R and RStudio have functionality for managing packages:
 > > ## Solution to challenge 5
 > >
 > > We can use the `install.packages()` command to install the required packages.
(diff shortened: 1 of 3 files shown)


naupaka:
THANKS @HLindsay! This is exactly the type of PR we like to see. Resolves three issues in one PR, to boot! And you are correct to just PR the modified `.Rmd` files. When I regenerate the site every few weeks, that will generate the `.md`, which will in turn get processed by Jekyll to make the html.

PR #233, opened 2016-12-07
add detail to why 'garbage row' appears

funkhou9:
When going through the lesson for the first time, when it came time to
remove a 'garbage row', I wasn't sure where it came from! I have tried
to add more explanation why the garbage row was created, when it was
created.

(SWC training checkout)


naupaka:
I might also add that the NA is only added to the factor columns in the data frame. The numeric columns still get appended to with the values from the list. Maybe add a sentence to this effect as well?

funkhou9:
Agreed naupaka. I tried to add some more discussion about why we only see NAs generated in the factor column, but not in the numeric or logical columns.

naupaka:
Thanks @funkhou9, looks good to me.

Issue #230, opened 2016-12-07
dplyr episode: incomplete solution for challenge 2


fmichonneau:
the solution gives the answer to the first part of the question but not the second (min and max values).

liz-is:
I'm happy to add a solution for this as part of completing instructor training. Would it be better to provide a solution that uses `arrange()` from dplyr, which has not been introduced in the lesson, or that uses `which.max()/which.min()` and subsetting the data frame, or would it be good to have both of these types of solutions? 

`arrange()` is the most dplyr-like way to do this but adds a function that learners won't have seen previously so I don't know if it's the best solution in this case.

liz-is:
For now I've added solutions using both approaches.

naupaka:
Hi @liz-is I've just added comments to your PR #236 that should clarify what you've asked here.

naupaka:
Closed by #236

PR #236, opened 2016-12-19
added two possible solutions to dplyr challenge 2

liz-is:
Added solutions for the second part of the challenge, as suggested in issue #230.

I've added solutions with and without the use of `arrange()`, as learners haven't seen this function during the lesson.

### DIFF
--- a/_episodes_rmd/13-dplyr.Rmd
+++ b/_episodes_rmd/13-dplyr.Rmd
@@ -181,6 +181,17 @@ even better.
 > >lifeExp_bycountry <- gapminder %>%
 > >    group_by(country) %>%
 > >    summarize(mean_lifeExp=mean(lifeExp))
+> >
+> >lifeExp_bycountry$country[which.max(lifeExp_bycountry$mean_lifeExp)]
+> >lifeExp_bycountry$country[which.min(lifeExp_bycountry$mean_lifeExp)]
+> >```
+> Another way to do this is to use the `dplyr` function `arrange()`, which has 
+> similar syntax to other functions from the `dplyr` package.
+> >```{r}
+> >lifeExp_bycountry %>%
+> >    arrange(mean_lifeExp)
+> >lifeExp_bycountry %>%
+> >    arrange(desc(mean_lifeExp))
 > >```
 > {: .solution}
 {: .challenge}
--- a/_episodes_rmd/13-dplyr.Rmd
+++ b/_episodes_rmd/13-dplyr.Rmd
@@ -173,25 +173,29 @@ even better.
 > ## Challenge 2
 >
 >
-> Calculate the average life expectancy per country. Which had the longest life
-> expectancy and which had the shortest life expectancy?
+> Calculate the average life expectancy per country. Which has the longest average life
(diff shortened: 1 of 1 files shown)


fmichonneau:
Personally, I think it would be best to stick to a dplyr approach for a solution to this challenge.
Another possibility would be to do: `... %>% filter(mean_lifeExp == min(mean_lifeExp) | mean_lifeExp == max(mean_lifeExp))`

_episodes_rmd/13-dplyr.Rmd
```diff
@@ -181,6 +181,17 @@ even better.
 > >lifeExp_bycountry <- gapminder %>%
 > >    group_by(country) %>%
 > >    summarize(mean_lifeExp=mean(lifeExp))
+> >
+> >lifeExp_bycountry$country[which.max(lifeExp_bycountry$mean_lifeExp)]
+> >lifeExp_bycountry$country[which.min(lifeExp_bycountry$mean_lifeExp)]
+> >```
+> Another way to do this is to use the `dplyr` function `arrange()`, which has
```
naupaka: I might add a bit on what `arrange()` does in this sentence.

naupaka: Also I might also might modify the challenge itself to clarify what is being asked for - is it the min/max individual value or the lowest/highest mean value that we'd like to find?

_episodes_rmd/13-dplyr.Rmd
```diff
@@ -181,6 +181,17 @@ even better.
...
 > >    summarize(mean_lifeExp=mean(lifeExp))
+> >
+> >lifeExp_bycountry$country[which.max(lifeExp_bycountry$mean_lifeExp)]
+> >lifeExp_bycountry$country[which.min(lifeExp_bycountry$mean_lifeExp)]
+> >```
+> Another way to do this is to use the `dplyr` function `arrange()`, which has 
+> similar syntax to other functions from the `dplyr` package.
+> >```{r}
+> >lifeExp_bycountry %>%
+> >    arrange(mean_lifeExp)
```
naupaka: Also could pipe into a `head(1)` call to each of these to get just the min and the max rows.

_episodes_rmd/13-dplyr.Rmd
```diff
@@ -181,6 +181,17 @@ even better.
 > >lifeExp_bycountry <- gapminder %>%
 > >    group_by(country) %>%
 > >    summarize(mean_lifeExp=mean(lifeExp))
+> >
+> >lifeExp_bycountry$country[which.max(lifeExp_bycountry$mean_lifeExp)]
+> >lifeExp_bycountry$country[which.min(lifeExp_bycountry$mean_lifeExp)]
```
naupaka: Might take these out since we're focusing on the dplyr approach in this episode.

liz-is:
I've made the changes you suggested, and added @fmichonneau's suggested solution as well. 

Do you think it would be better to move the information about `arrange()` to the start of the challenge, or is it okay where it is?

naupaka:
I think it looks good as is. Thanks for making the changes so quickly. Addresses #230

PR #237, opened 2016-12-21
Lesson 05 - extend objectives + add headers

ktoddbrown:
+Extended the objectives to reflect specific lesson content.
+Added headers to subset material
+Removed redundant examples

### DIFF
--- a/_episodes_rmd/05-data-structures-part2.Rmd
+++ b/_episodes_rmd/05-data-structures-part2.Rmd
@@ -3,22 +3,22 @@ title: "Exploring Data Frames"
 teaching: 20
 exercises: 10
 questions:
-- "How can I manipulate a dataframe?"
+- "How can I manipulate a data frame?"
 objectives:
 - "Be able to add and remove rows and columns."
 - "Be able to remove rows with `na` values."
-- "Be able to append two data.frame"
+- "Be able to append two data frames"
 - "Be able to articulate what a `factor` is and how to convert between `factor` and `character`."
-- "Be able to find basic properties of a data.frame including size, class or type of the columns, names, and first few rows."
+- "Be able to find basic properties of a data frames including size, class or type of the columns, names, and first few rows."
 keypoints:
-- "Use `cbind()` to add a new column to a data.frame."
-- "Use `rbind()` to add a new row to a data.frame."
-- "Remove rows from a dataframe."
-- "Use `na.omit()` to remove rows from a data.frame with `NA` values."
+- "Use `cbind()` to add a new column to a data frame."
+- "Use `rbind()` to add a new row to a data frame."
+- "Remove rows from a data frame."
+- "Use `na.omit()` to remove rows from a data frame with `NA` values."
 - "Use `levels()` and `as.character()` to explore and manipulate factors"
-- "Use `str()`, `nrow()`, `ncol()`, `dim()`, `colnames()`, `rownames()`, `head()` and `typeof()` to understand structure of the data.frame"
+- "Use `str()`, `nrow()`, `ncol()`, `dim()`, `colnames()`, `rownames()`, `head()` and `typeof()` to understand structure of the data frame"
 - "Read in a csv file using `read.csv()`"
-- "Understand `length()` of a data.frame"
+- "Understand `length()` of a data frame"
 ---
 
 ```{r, include=FALSE}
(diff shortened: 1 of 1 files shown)


_episodes_rmd/05-data-structures-part2.Rmd
```diff
@@ -27,6 +34,8 @@ the data.frame - that table that we started with that information from a CSV
 gets dumped into when we load it. In this lesson, we'll learn a few more things
 about working with data.frame.
 
+##Adding columns and rows in data.frame
```
naupaka: I believe our current style guide (I *think* we have a style guide?) suggests a space after the pound signs for headers. This is how the other lessons are. Can you add a space after each pair of `##` before the heading text in the heading throughout?

_episodes_rmd/05-data-structures-part2.Rmd
```diff
@@ -5,13 +5,20 @@ exercises: 10
 questions:
 - "How can I manipulate a dataframe?"
 objectives:
-- "To learn how to manipulate a data.frame in memory"
-- "To tour some best practices of exploring and understanding a data frame when it is first loaded."
+- "Be able to add and remove rows and columns."
+- "Be able to remove rows with `na` values."
```
naupaka: should be `NA` for consistency

_episodes_rmd/05-data-structures-part2.Rmd
```diff
@@ -5,13 +5,20 @@ exercises: 10
 questions:
 - "How can I manipulate a dataframe?"
```
naupaka: I think this is as good a place as any to start fixing the inconsistent use of `data.frame` and data frame throughout these lessons. We had a long discussion this a while back on #145 and I think the conclusion was to use "data frame" when referring to the thing and `data.frame` only when referring to the function `data.frame()`.  So in that vein, could you change the references to the thing throughout to "data frame" instead of as above "dataframe" or as you have below `data.frame`. One of these days I'm going to go through and do the same thing for all the other lessons as well.

_episodes_rmd/05-data-structures-part2.Rmd
```diff
@@ -5,13 +5,20 @@ exercises: 10
...
 objectives:
-- "To learn how to manipulate a data.frame in memory"
-- "To tour some best practices of exploring and understanding a data frame when it is first loaded."
+- "Be able to add and remove rows and columns."
+- "Be able to remove rows with `na` values."
+- "Be able to append two data.frame"
```
naupaka: The shift to "data frame" also helps here, because then you can just say "data frames".

_episodes_rmd/05-data-structures-part2.Rmd
```diff
@@ -168,6 +183,7 @@ cats
 > {: .solution}
 {: .challenge}
 
+#Realistic example
```
naupaka: Should be a second-level header, no?

_episodes_rmd/05-data-structures-part2.Rmd
```diff
@@ -208,8 +224,7 @@ We can also examine individual columns of the data.frame with our `typeof` funct
 
 ```{r}
 typeof(gapminder$year)
-typeof(gapminder$lifeExp)
-typeof(gapminder$country)
```
naupaka: I think the idea with having several examples is that they show different types. Perhaps at least leave `typeof(gapminder$country)` in to show the difference between that and the year column?

_episodes_rmd/05-data-structures-part2.Rmd
ktoddbrown: I've updated the data frame references.

_episodes_rmd/05-data-structures-part2.Rmd
ktoddbrown: Done.

_episodes_rmd/05-data-structures-part2.Rmd
ktoddbrown: Done

_episodes_rmd/05-data-structures-part2.Rmd
ktoddbrown: If we have a style guide I would love a link :) But yes done.

_episodes_rmd/05-data-structures-part2.Rmd
ktoddbrown: I propose leaving this as a first-level header because it summarizes the entire lesson previously. No changes made.

_episodes_rmd/05-data-structures-part2.Rmd
ktoddbrown: I'm not a huge fan but I can see your point and put this back in.

_episodes_rmd/05-data-structures-part2.Rmd
naupaka: We generally follow the google R style guide for the SWC/DC R materials, so I suppose the google markdown style guide is also something to shoot for: https://github.com/google/styleguide/blob/gh-pages/docguide/style.md

_episodes_rmd/05-data-structures-part2.Rmd
naupaka: A agree with your point about the last example being a higher-level summary, but the H2-only convention used throughout all of the SWC lessons is based on how Jekyll and our CSS format the lesson; we ask folks not to use nested headers for this reason. See [here on the lesson-formatting page](http://swcarpentry.github.io/lesson-example/04-formatting/) for more details.

ktoddbrown: Done. Sorry about that.

naupaka:
Thanks!

PR #238, opened 2016-12-23
Added a clarification to lesson 10-functions that the variables in th…

robertbagchi:
…e function are passed as arguments - this might not have been clear to everyone before


naupaka:
Thanks!

PR #240, opened 2017-01-09
Episode 3, challenge 1: trailing backtick breaks example 

pboesu:
There is a trailing backtick in the third example vector of challenge 1 in episode 3. On verbatim execution in R this prohibits the correct execution of the c() function, instead leading to an R prompt expecting further input.


naupaka:
Thanks for this! Not sure how it snuck in there.

PR #245, opened 2017-02-16
Instructor-training assignment

jordwil:
Added an additional feature to the challenge, mainly to introduce reserved words as well as add descriptions to the solutions.

### DIFF
--- a/_episodes/01-rstudio-intro.md
+++ b/_episodes/01-rstudio-intro.md
@@ -761,12 +761,17 @@ network). R and RStudio have functionality for managing packages:
 > > The following will not be able to be used to create a variable
 > > 
 > > ~~~
-> > _age - We can only start variable names with letters or dots not followed by a number.
-> > min-length - This contains a dash "-" character that can't be used when naming a variable. We can only use letters,
-> > numbers, dots, and underline "_" characters.
-> > 2widths - 2widths - We can't start a variable name with a number.
-> > if - If is a special case called a reserved word. We use it for control structures that will be discussed at a later
-> > date.
+> > _age 
+> >     * We can only start variable names with letters or dots not followed by a number.
+> > min-length
+> >     * We can only use letters, numbers, dots, and underline "_" characters to assign our variable.
+> > 2widths 
+> >     * We can't start our assignment with a number.
+> > if 
+> >     * 'if' is considered a reserved word. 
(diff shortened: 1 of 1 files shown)


_episodes/01-rstudio-intro.md
```diff
@@ -760,9 +761,12 @@ network). R and RStudio have functionality for managing packages:
 > > The following will not be able to be used to create a variable
 > > 
 > > ~~~
-> > _age
-> > min-length
-> > 2widths
+> > _age - We can only start variable names with letters or dots not followed by a number.
+> > min-length - This contains a dash "-" character that can't be used when naming a variable. We can only use letters,
+> > numbers, dots, and underline "_" characters.
+> > 2widths - 2widths - We can't start a variable name with a number.
```
naupaka: Why is '2widths' repeated?

_episodes/01-rstudio-intro.md
```diff
@@ -760,9 +761,12 @@ network). R and RStudio have functionality for managing packages:
...
+> > if - If is a special case called a reserved word. We use it for control structures that will be discussed at a later
```
naupaka: I might say something like: "`If` is a special type of word called a reserved word. There are about 20 of these reserved words, and they represent things that cannot be directly overwritten during normal use of the assignment operator because they are a core part of the language itself.

naupaka:
You should also give the PR a descriptive name that says what it does to the lesson. Then in the description text you can note that it is being submitted as part of instructor training.

naupaka:
Hi @jordwil thanks for the modifications. I didn't notice this the first time, but your PR contains modifications to the `.md` file for the lesson and not the `.Rmd`. Since we generate the `.md` files from the `.Rmd` files, the plain markdown will get overwritten whenever we rebuild the site content. The guide for contributions to lessons is [here](https://swcarpentry.github.io/lesson-example/). It will be cleaner and potentially easier if we close this PR and you reopen a new one with just modifications to the corresponding `.Rmd` file in the `_episodes_rmd` directory. Let me know if that sounds reasonable.

naupaka:
Hi @jordwil - do you think you'll have a chance to migrate these changes over to the `Rmd` instead of the `md` files? Otherwise I'll close this PR since there hasn't been much activity for a few months.

jordwil:
I can do this tomorrow.

PR #247, opened 2017-02-27
Modified illustration example to fit previous example better.

ytakemon:
When I was going over lesson 06 for subsetting data, I got a little confused by why the example illustration jumped back to the initial x vector assignment c("a", "b", "c", "d", "e"). I thought it would be better for continuity if the example illustration used the x vector previously assigned c("a", "a", "a") and went over what happens with you %in% with c("a", "c").


naupaka:
Hi @ytakemon thanks for the PR, and sorry for the wait on getting feedback. I agree that switch to a different vector was confusing, and I think your changes help smooth it over. Thanks for adding the bit about vector recycling as well.

PR #249, opened 2017-02-28
Fix L suffix as proposed by @rgaiacs

### LINKED ISSUE #192: L suffix
Can we move the note

> Note the L suffix to insist that a number is an integer.

to be a comment like

> ```
> typeof(1L)  # The L suffix is used to insist that a number is an integer since by default R use float numbers.
> ```

marwahaha:
Fixes #192


naupaka:
Thanks @marwahaha

PR #254, opened 2017-03-13
Quoted all calls to library() and added [R for Data Science](r4ds.had.co.nz) links

martinjhnhadley:
Also took opportunity to remove quotes in library calls, for consistency.
And a link to r4ds.had.co.nz

### DIFF
--- a/_episodes/08-plot-ggplot2.md
+++ b/_episodes/08-plot-ggplot2.md
@@ -43,8 +43,7 @@ Let's start off with an example:
 
 
 ~~~
-library(ggplot2)
-library(gapminder)
+library("ggplot2")
 ggplot(data = gapminder, aes(x = gdpPercap, y = lifeExp)) +
   geom_point()
 ~~~
--- a/_episodes/08-plot-ggplot2.md
+++ b/_episodes/08-plot-ggplot2.md
@@ -43,7 +43,8 @@ Let's start off with an example:
 
 
 ~~~
-library("ggplot2")
+library(ggplot2)
+library(gapminder)
 ggplot(data = gapminder, aes(x = gdpPercap, y = lifeExp)) +
   geom_point()
 ~~~
--- a/_episodes/14-tidyr.md
+++ b/_episodes/14-tidyr.md
@@ -56,8 +56,8 @@ Load the packages
 
 
 ~~~
-library("tidyr")
-library("dplyr")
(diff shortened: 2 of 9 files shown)


naupaka:
Hi @martinjhnhadley thanks for the PR, and the addition of Hadley's new book to the links. 

However, you've edited the markdown and not the Rmd files. The markdown files are generated from the Rmd every time I rebuild the site, so they'll get overwritten. If you'd like to open up a new PR to add the link to R4DS to the corresponding Rmd file, I'd be happy to add that in.

Additionally, I'd like for the lessons to all use quoted names for loading packages instead of the unquoted NSE way of doing it. For the rationale, see [here](http://stackoverflow.com/questions/36802132/the-use-of-quotation-marks-when-loading-a-package-in-r) and [here](http://adv-r.had.co.nz/Computing-on-the-language.html#nse-downsides). If calls to `library()` lack quotes elsewhere in the lessons, then I see that as an issue that should be fixed. 

And finally, we don't need to load the `gapminder` package in these lessons, since they are loaded from a csv at the top of the lessons (e.g. the [ggplot2 lesson](https://github.com/swcarpentry/r-novice-gapminder/blob/gh-pages/_episodes_rmd/08-plot-ggplot2.Rmd)). There was discussion about loading from a csv or from the `gapminder` package, [here](https://github.com/swcarpentry/r-novice-gapminder/issues/172), and we decided to go with the csv to lower the number of required packages for the lessons, and to prevent things breaking if the data were to be updated elsewhere.

martinjhnhadley:
@naupaka thank you for your thorough reply!
I'll update this PR with changes to the Rmds in the morning (GMT), switching unquoted library calls per the sensible discussions you linked.
I should've checked the issues for gapminder, the issue you linked raises good points.
Cheers,
Martin

naupaka:
Thanks for the fixes. Just to keep things tidy, can you also do a `git revert` on your commit f46f5b003ca ? Sometime the makefile gets confused when people modify the `md` files directly.

martinjhnhadley:
@naupaka done but perhaps not in the most elegant manner

naupaka:
Hrm... I'm not sure what you did. It looks like you removed and then added the changes back in? I'm a little confused. I will merge and fix it, though. No worries.

Thanks for the PR, @martinjhnhadley!

PR #257, opened 2017-04-10
Proof and copy edit for 04-data-structures-part1.Rmd

DanBuchan:
For my instructor checkout I've given this lesson a basic proof and copy edit. It includes a handful of grammatical or styles changes to make it read more smoothly and correct one or two typos. The section on read.table/read.delim/read.csv has been rewritten.

Change notes by line:

35: "like what you might already have" not very natural english, not typically regarded as valid for (formal) written text
66: Updated the read.table/read.csv/read.delim description to correctly reflect who and why R provides these functions
156: "When R reads a csv file" - corrected the article
231: removed double 'that' for clearer reading
245: grammatical change to "such as" to clarify example is being given
310: moved c() to it's own clause and made explicit that we are talking about a function.
404: Clarified why R represents factors as indices

### DIFF
--- a/_episodes_rmd/04-data-structures-part1.Rmd
+++ b/_episodes_rmd/04-data-structures-part1.Rmd
@@ -62,15 +62,15 @@ cats <- read.csv(file = "data/feline-data.csv")
 cats
 ```
 
-
 The `read.table` function is used for reading in tabular data stored in a text
-file where the columns of data are delimited by punctuation characters such as
+file where the columns of data are separated by punctuation characters such as
 CSV files (csv = comma-separated values). Tabs and commas are the most common
-punctuation characters used to delimit data points in csv files. For
-convenience R provides 2 other versions of `read.table`. These are `read.csv`
-for files where the data is separated with commas and `read.delim` for files
-where the data is separated with tabs. If needed it is possible to override the
-default punctuation marks for both `read.csv` and `read.delim`.
+punctuation characters used to separate or delimit data points in csv files. 
+For convenience R provides 2 other versions of `read.table`. These are: `read.csv`
+for files where the data are separated with commas and `read.delim` for files
+where the data are separated with tabs. Of these three functions `read.csv` is
+the most commonly used.  If needed it is possible to override the default 
+delimiting punctuation marks for both `read.csv` and `read.delim`.
 
 
 We can begin exploring our dataset right away, pulling out columns by specifying
@@ -228,7 +228,7 @@ empty character strings. If we similarly do
 str(cats$weight)
 ```
 
-we see that is a vector, too - *the columns of data we load into R
+we see that `cats$weight` is a vector, too - *the columns of data we load into R
 data.frames are all vectors*, and that's the root of why R forces everything in
 a column to be the same basic data type.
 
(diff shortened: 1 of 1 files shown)


_episodes_rmd/04-data-structures-part1.Rmd
```diff
@@ -63,12 +63,14 @@ cats
...
-this format you can use the function `read.delim`. If the columns in your data
-are delimited by a character other than commas or tabs, you can use the more
-general and flexible `read.table` function.
+The `read.table` function is used for reading in tabular data stored in a text
+file where the columns of data are delimited by punctuation characters such as
+CSV files (csv = comma-separated values). Tabs and commas are the most common
+punctuation characters used to delimit data points in csv files. For
+convenience R provides 2 other versions of `read.table`. These are `read.csv`
+for files where the data is separated with commas and `read.delim` for files
+where the data is separated with tabs. If needed it is possible to override the
```
naupaka: I think in this case it would be clearer to use 'data' as plural, so 'data ia' becomes 'data are'.

naupaka: I also see why you revised it in this way, but I also understand why it was originally written the way it was — most people will meet `read.csv()` first, and only later learn/realize that it is a special case of `read.table()`. Perhaps the middle ground is to add a sentence to the effect  of `read.csv()` being the function most commonly used?

_episodes_rmd/04-data-structures-part1.Rmd
```diff
@@ -227,7 +228,7 @@ empty character strings. If we similarly do
 str(cats$weight)
 ```
 
-we see that that's a vector, too - *the columns of data we load into R
+we see that is a vector, too - *the columns of data we load into R
```
naupaka: I would change to "we see that `cats$weight` is a vector, too" to clarify further. This also helps to avoid the double "that".

_episodes_rmd/04-data-structures-part1.Rmd
```diff
@@ -241,7 +242,7 @@ a column to be the same basic data type.
 > > By keeping everything in a column the same, we allow ourselves to make simple
 > > assumptions about our data; if you can interpret one entry in the column as a
 > > number, then you can interpret *all* of them as numbers, so we don't have to
-> > check every time. This consistency, like consistently using the same separator
+> > check every time. This consistency, such as consistently using the same separator
 > > in our data files, is what people mean when they talk about *clean data*; in
```
naupaka: I might just delete this clause ("such as consistently using the same in our data files") altogether. Using a tsv and a csv in the same project doesn't mean your data are dirty.

_episodes_rmd/04-data-structures-part1.Rmd
DanBuchan: As suggested I've change the pluralisation and fixed the grammar there so it reads correctly.

I've added a sentence to note that read.csv is the most common.

Additionally on reading I realised novices likely haven't encountered the word delimited before. So I replaced the first instance with 'separated' and then added 'delimit' as an alternative for separate later.

_episodes_rmd/04-data-structures-part1.Rmd
DanBuchan: added

_episodes_rmd/04-data-structures-part1.Rmd
DanBuchan: removed

naupaka:
Thanks!

PR #258, opened 2017-04-11
Suggested edits to code and explanation for setting labels in ggplot

orb16:
Under the section ## Modifying text heading. Using the `labs` argument might be simpler to explain `xlab`, `ylab`, `ggtitle`, and putting the header in an otherwise unnecessary `scale_colour` as in the example?

Sorry for the double commit, hadn't deleted the scale_colour_discrete in the first one.

### DIFF
--- a/_episodes_rmd/08-plot-ggplot2.Rmd
+++ b/_episodes_rmd/08-plot-ggplot2.Rmd
@@ -331,14 +331,13 @@ elements. The x-axis is too cluttered, and the y axis should read
 "Life expectancy", rather than the column name in the data frame.
 
 We can do this by adding a couple of different layers. The **theme** layer
-controls the axis text, and overall text size, and there are special layers
-for changing the axis labels. To change the legend title, we need to use the
-**scales** layer.
+controls the axis text, and overall text size. Labels for the axes, plot title and legend can be set using the `labs` function.
 
 ```{r theme}
 ggplot(data = az.countries, aes(x = year, y = lifeExp, color=continent)) +
   geom_line() + facet_wrap( ~ country) +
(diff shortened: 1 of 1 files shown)


_episodes_rmd/08-plot-ggplot2.Rmd
```diff
@@ -331,15 +331,13 @@ elements. The x-axis is too cluttered, and the y axis should read
...
-**scales** layer.
+controls the axis text, and overall text size. Labels for the axes, plot title and legend can be set using the `labs` function. 
 
 ```{r theme}
 ggplot(data = az.countries, aes(x = year, y = lifeExp, color=continent)) +
   geom_line() + facet_wrap( ~ country) +
-  xlab("Year") + ylab("Life expectancy") + ggtitle("Figure 1") +
-  scale_colour_discrete(name="Continent") +
+  labs(x = "Year", y = "Life expectancy", 
+       title = "Figure 1", colour = "Continent") +
```
naupaka: Two thoughts: one is that since you are already revising this, to also add more spacing, so each of the parameters set in `labs()` is on it's own line. Makes it easier to read and to refer to on screen. The second is that it is not immediately obvious why setting the 'colour' parameter changes the name of the legend. (Admittedly also not super intuitive that you can do this with the `scale_colour_discrete()` function, but so it goes). Can you add a few words to your description on line 334 to point out explicitly that this is how you set set the legend title?

naupaka:
Thanks for the PR. Just made a few small suggestions. I have always set these labels using different layers, but I can see how putting them all in together with `labs()` is a bit more intuitive to write and to read.

orb16:
Updated with https://github.com/swcarpentry/r-novice-gapminder/pull/277 as mistakenly deleted by original fork - should address @naupaka's suggestions

PR #260, opened 2017-04-19
Add template language for issues and PRs about lesson contribution an…

ErinBecker:
…d maintenance process

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,9 @@
+Please delete the text below before submitting your contribution. 
+
+---
+
+Thanks for contributing! If this contribution is for instructor training, please send an email to checkout@carpentries.org with a link to this contribution so we can record your progress. You’ve completed your contribution step for instructor checkout just by submitting this contribution.  
+
+Please keep in mind that lesson maintainers are volunteers and it may be some time before they can respond to your contribution. Although not all contributions can be incorporated into the lesson materials, we appreciate your time and effort to improve the curriculum.  If you have any questions about the lesson maintenance process or would like to volunteer your time as a contribution reviewer, please contact Kate Hertweck (k8hertweck@gmail.com).  
+
+---
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,9 @@
+Please delete the text below before submitting your contribution. 
+
+---
+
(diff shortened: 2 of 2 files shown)


naupaka:
Didn't it used to be that PRs had to be merged to complete checkout? I guess that doesn't scale with the larger numbers of folks being trained?

naupaka:
Also - should we continue to tag checkout contributions with the 'instructor-training' label? @gvwilson was doing this for a while.

gvwilson:
I found it helpful when prioritizing material for review.

naupaka:
@gvwilson but it doesn't tie into some fancy behind-the-scenes script magic thingy?

gvwilson:
Nope, but I did have some canned GitHub queries to pull up all the 
issues in swcarpentry repos with that tag.

ErinBecker:
@naupaka - PRs no longer need to be merged for checkout (and haven't for at least the past year, since I started). To my knowledge the "instructor-training" label has been helpful in identifying which PRs/issues are from trainees (both for prioritizing review and recording progress). In terms of recording progress, the tag is no longer needed, due to other process changes, but if still helpful for prioritizing review by all means keep using!

naupaka:
@ErinBecker thanks for the clarification :)

So for old PRs where recommended changes are not made, I should check in with the submitter and then feel (more than before) ok closing them without merging? Less incentive for submitter to follow up if no requirement to have it merged.

ErinBecker:
@naupaka This raises a more complicated question of what our procedures around reviewing PRs should be (e.g. how long should we wait before closing stale PRs? If there is community interest in seeing a change made, but the original person who raised the issue doesn't have the technical expertise (or time) to implement the change, what should a maintainer do? among other questions). 

I don't have an answer to those questions right now, but I'm chatting with @k8hertweck next week about guidelines for maintainers around these questions. Do you have any recommended points to make sure we hit upon?

naupaka:
It agree it would be great to have publicly posted guidelines on some of these things. Definitely one would be how long we should keep stale PRs around - everyone is busy, so not everything happens overnight, but some can hang around for years. Same thing with issues - if someone adds an issue, but there is no discussion about it and no PR to fix it is ever submitted, how long do we wait before closing it? Seems important to balance not having old cruft hanging around with being welcoming to people new to contributing.

ErinBecker:
Agreed on all counts. Often I think what ends up happening is that issues stick around for months (or longer) with no activity and then are closed without warning. Having posted guidelines about this process will help make the whole process more clear and transparent for everyone (and the process less frustrating for both maintainers and contributors).

PR #261, opened 2017-04-20
Replace list example title

lonsbio:
'Research Bazaar' seems a holdover from previous lesson contributors, changed to be more generic.


naupaka:
Thanks!

Issue #256, opened 2017-03-31
"Seeking Help" challenge 2 answer & example misleading


delocalizer:
The explanation of the difference between `sep` and `collapse` is if not wrong, then pretty misleading - the actual difference between these is that `sep` string separates _terms,_ and `collapse` joins _results_ after concatenation. The example given only works because the arguments supplied are of length 1, which is a sort of degenerate case for `paste`. A clearer example would be: 
```
> paste(1:2, 3:4)
[1] "1 3" "2 4"
> paste(1:2, 3:4, sep = ',')
[1] "1,3" "2,4"
> paste(1:2, 3:4, sep = ',', collapse = '|')
[1] "1,3|2,4"
```
but this presumes a knowledge of vectors that the lesson hasn't yet reached at this point. So perhaps choose a different function than `paste` to use as an example?

naupaka:
@delocalizer thanks for the comment. Which episode and line number is the current explanation on?

delocalizer:
Hi @naupaka, it is episode 3, in challenge 2:
https://github.com/swcarpentry/r-novice-gapminder/blame/gh-pages/_episodes/03-seeking-help.md#L177

naupaka:
@delocalizer I agree the current explanation is not so helpful. I also agree your example is clearer, but I think you're right about the problem about relying on a solid understanding of vectorized operations.  

What about:
```r
> paste(c("a","b"), "c")
[1] "a c" "b c"
> paste(c("a","b"), "c", sep = ",")
[1] "a,c" "b,c"
> paste(c("a","b"), "c", sep = ",", collapse = "|")
[1] "a,c|b,c"
```

delocalizer:
Yes, that looks good to me

naupaka:
@delocalizer want to submit a PR with this change?

delocalizer:
sure thing


________________________________

naupaka:
Closed by #262

PR #262, opened 2017-04-27
clearer explanation of difference between sep and collapse

delocalizer:
Addresses #256

### DIFF
--- a/_episodes/03-seeking-help.md
+++ b/_episodes/03-seeking-help.md
@@ -171,13 +171,25 @@ your issue.
 > > ?paste
 > > ~~~
 > > {: .r}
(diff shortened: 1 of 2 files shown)


naupaka:
Thanks for the PR @delocalizer - but with the R lessons, the source files to modify are the `.Rmd` files, in this case [this one](https://github.com/swcarpentry/r-novice-gapminder/blob/gh-pages/_episodes_rmd/03-seeking-help.Rmd). The `.md` files get regenerated every time I rebuild all the content, so any changes made to them get overwritten. Can you apply your PR to the `Rmd` file instead?

delocalizer:
of course. Sorry about that.

naupaka:
Thanks!

PR #265, opened 2017-05-02
Fix link to R for Data Science in dplyr lesson

jcoliver:
Link to http://r4ds.had.co.nz/ in the dplyr lesson (episode 13) was incomplete.


naupaka:
Thanks!

PR #269, opened 2017-05-08
s/scatter/spread as keypoint

lwjohnst86:
The keypoint says "scatter() to convert from long to wide" but there is no scatter command. It should be spread


naupaka:
Wow - that's a good catch of an obvious mistake. Thanks!

PR #279, opened 2017-06-06
Update 03-seeking-help.Rmd

hkronenb:
I made a small changer here because I feel like "<-" is less intuitive to a new R user than "+" and  therefore might be a better place to start when showing help examples.


naupaka:
Thanks!

PR #281, opened 2017-06-26
Update 05-data-structures-part2.Rmd Challenge 2

ReneHeim:
I would propose to change Challenge 2 because is has no relation to the section that is leading to this challenge. The mentioned challenge asks to use the source function but this function has not been mentioned before. I would either explain why source() is used at all and provide an example (tiny function maybe that is saved as a script somewhere and is then sourced) or I would not cover source in this lesson and rather offer a challenge that shows the student maybe some in-built datasets and (e.g. Iris). The student could add another row that contains useful calculations from already extisting rows (e.g. standard deviation). Such a task would further familiarize the student with manipulating dataframes by using an example that could show up in reality as well.

---

---

### DIFF
--- a/_episodes_rmd/05-data-structures-part2.Rmd
+++ b/_episodes_rmd/05-data-structures-part2.Rmd
@@ -291,6 +291,15 @@ into a script file so we can come back to it later.
 > Run the script using the `source` function, using the file path
 > as its argument (or by pressing the "source" button in RStudio).
 >
+> The 'source' function can be used to use a script within a script.
+> Assume you would like to load the same type of file over and over 
+> again and therefore you need to specify the arguments to fit the 
+> needs of your file. Instead of writing the necessary argument again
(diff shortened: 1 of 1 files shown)


naupaka:
I agree with your critique @ReneHeim. I'm not sure how other people feel about this, but my sense is that the `source()` function becomes most useful once you've learned about writing functions and can move those functions to a separate file to keep your main script more tidy. So maybe this content belongs somewhere over near where we cover functions?

tomwright01:
I often use source commands to seperate (and hide) complex chunks of code, as such I do often use it to load and clean my data. However I agree it's really poorly integrated in it's current location. 
Perhaps it would fit best before __combining function__ in `10-functions` , we could then demonstrate how a source file can be used to define some utility functions `fahr_to_kelvin` and `kelvin_to_celcius` which can then be used in the __combining functions__ section.

Issue #178, opened 2016-09-02
Environment doesn't have y as variable


rgaiacs:
> Managing your environment
> 
> There are a few useful commands you can use to interact with the R session.
> 
> `ls` will list all of the variables and functions stored in the global environment (your working R session):
> 
> ```
> ls()
> [1] "x" "y"
> ```

but we never defined `y` before.

naupaka:
What lesson and line is this from?

liz-is:
This occurs in [lesson1, line 389](https://github.com/swcarpentry/r-novice-gapminder/blob/gh-pages/_episodes_rmd/01-rstudio-intro.Rmd#L389).

naupaka:
It's because of the script used to build all the lessons uses `x, y` as parameter values, so then these exist as objects inside the function for all of the lessons. See: https://github.com/swcarpentry/r-novice-gapminder/blame/gh-pages/bin/generate_md_episodes.R#L30

Example of the situation:

```r
test <- function(x, y) {
  # this is where all the knit() calls go
  print(ls())
}
test(1,2)
[1] "x" "y"
```

Not sure best way to get around this. Maybe easiest (although hacky) would be to just assign something to `y` so that the output of `ls()` isn't surprising.

naupaka:
Closed by #282

PR #282, opened 2017-06-27
define variable 'y'

vince-p:
The output of the command ls() on line 397 shows two variables, x and y. But (confusingly) there is no place that the y variable is defined in the current lesson. This simple change will make the output of ls() consistent for anyone who is following along exactly.


naupaka:
Thanks for the PR. This problem is described in #178. The root of it is the function that renders the `.Rmd`, but I think this fix is the most simple and straightforward way to handle the issue. Closes #178.

PR #283, opened 2017-06-29
Fix natural log bug.

danielsmaxwell:
Fixed a minor bug in the 01_rstudio_intro.Rmd and .md files.  See my commit descriptions for details.

### DIFF
--- a/_episodes/01-rstudio-intro.md
+++ b/_episodes/01-rstudio-intro.md
@@ -285,7 +285,7 @@ sin(1)  # trigonometry functions
 
 
 ~~~
-log(1)  # natural logarithm
+exp(1)  # natural logarithm
 ~~~
 {: .r}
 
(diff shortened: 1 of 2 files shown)


naupaka:
Hi @danielsmaxwell thanks for the PR. I think, however, that the code is correct as it is currently. The natural log of 1 is 0, not e. And `?log` shows that the default base for `log()` without the `base` parameter is e. 

Also, for future PRs, no need to edit the `.md` files, these are generated from the `.Rmd` when I rebuild the site. 

And I think you may have the incorrect line endings set with your git install, which is why the diff presents as the versions being completely different. They should be Linux-style line endings (`\n`). 

I'm going to close this PR, but if I've misunderstood something, we can reopen it and go from there.

danielsmaxwell:
Thanks for the update.  Apparently, I misunderstood the comment in the code as I thought that we wanted to show how to calculate the natural logarithm (2.718) in R rather than calculating the natural logarithm of 1 which is -- as you correctly pointed out -- 0.

Issue #275, opened 2017-05-30
Project Management With RStudio: Git


kekoziar:
One of the objectives of this episode is to use git from within RStudio, but there is no indication in the setup for the lesson that git should be installed, and it does not install automatically with R or RStudio.  If git is part of the lesson, installation and configuration should be required as part of the setup.  Perhaps the git parts of the lesson can be indicated as optional, with the setup having the same indication along with the link to git installation instructions?

naupaka:
I agree that this is a problem. 

I think it potentially makes sense to remove that git section altogether, since it is cursory and doesn't have the necessary background included. Plus, there is a far better version of this content [over in the git-novice lessons](http://swcarpentry.github.io/git-novice/14-supplemental-rstudio/). We could just link to that.

My inclination would be to remove that whole section in that episode and just link out to the other content as a callout. What do you think @kekoziar?

kekoziar:
I agree that linking to other content would be a good solution.  I like the challenge associated with git, so I will can put an issue in at the git-novice lesson to add the challenge to their page.

naupaka:
Would you also be willing to submit a PR on this repo removing the current git content and adding that link?

kekoziar:
Yes.

naupaka:
@kekoziar still willing to submit that PR?

kekoziar:
@naupaka Yes.  Sorry, for the delay; I've been traveling for a bit.

naupaka:
Closed by #291

PR #291, opened 2017-07-25
Git Removal and Add Link

kekoziar:
Information on installing and configuring Git with R Studio is missing.  Since Git is a cursory part of this lesson, this proposed file change will remove Git references and replace with link to better version of content in SWC Git lesson.
https://github.com/swcarpentry/r-novice-gapminder/issues/275


naupaka:
Great, thanks!

PR #295, opened 2017-08-01
Fix interactive display for challenge 3

iglpdc:
The closing `{: .solution}` for challenge 3 is missing a `.`. This breaks the interactive display of the solution. I'm adding the missing `.`.


naupaka:
Thanks!

Issue #266, opened 2017-05-04
Factors, numbers and coercion


rgaiacs:
Today I was playing with R and I discovery that some times transform factors and numbers strait forward can not work as you expect:

~~~
> f <- factor(c(50, 30, 10, 90, 40, 30, 50, 60, 80, 30))
> f
 [1] 50 30 10 90 40 30 50 60 80 30
Levels: 10 30 40 50 60 80 90
> as.numeric(f)
 [1] 4 2 1 7 3 2 4 5 6 2
~~~

You need to use 

~~~
> as.numeric(as.character(f))
 [1] 50 30 10 90 40 30 50 60 80 30
~~~

I think this could be a good challenge or a good call out box since in a quick look I didn't found it on the lesson.

@swcarpentry/r-maintainers please tag this one as suitable for instructor training.

naupaka:
If it gets added, should probably be around [line 104 in Data Structures part 2](https://github.com/swcarpentry/r-novice-gapminder/blame/gh-pages/_episodes_rmd/05-data-structures-part2.Rmd#L104).

naupaka:
@rgaiacs I've only be using the instructor-training label for PRs submitted as part of instructor training. I've added 'help-wanted' and 'newcomer friendly' to this one.

adsteen:
I'm having trouble understanding the workflow for contributing (see issue #298) but I've emailed a .Rmd of lesson 05 that includes a challenge related to this problem. The updated .Rmd is also attached here (as a .txt since github doesn't support .Rmd). 
[05-data-structures-part2.txt](https://github.com/swcarpentry/r-novice-gapminder/files/1200904/05-data-structures-part2.txt)

naupaka:
Hi @adsteen thanks for the comment and the change! I noticed that you are scheduled to be in the vegan workshop at ESA this Sunday that I'll be teaching - perhaps we can take a few moments to go through the contribution workflow for lessons? Might be easier than online...

naupaka:
Resolved by #299

Issue #298, opened 2017-08-03
URLs for contributing in /blob/gh-pages/CONTRIBUTING.md


adsteen:
in `/blob/gh-pages/CONTRIBUTING.md` lines 48-50, the URLs given to work in for people who want to change the lesson are [https://github.com/swcarpentry/FIXME](https://github.com/swcarpentry/FIXME) and [https://swcarpentry.github.io/FIXME](https://swcarpentry.github.io/FIXME). These both give a 404 Not Found error.

naupaka:
Looks like this would be an easy thing to fix with a PR. See the equivalent fix over at r-inflammation: https://github.com/swcarpentry/r-novice-inflammation/pull/291

naupaka:
Resolves by #299

PR #299, opened 2017-08-10
added as.numeric(factor) challenge

adsteen:
I've added a challenge to lesson 05 in response to [issue 266](https://github.com/swcarpentry/r-novice-gapminder/issues/266)


naupaka:
Looks good! Thanks for the PR. Resolves #266 and #298

Issue #301, opened 2017-08-16
Inconsistent advice on where to write R code in Lessons 1 and 2


rlbarter:
In Lesson 1 the author wrote: 
> "Test and play within the interactive R console then copy code into a .R file to run later.
> - This works well when doing small tests and initially starting off.
> - It quickly becomes laborious
>
> Start writing in an .R file and use RStudio’s command / short cut to push current line, selected lines or modified lines to the interactive R console.
> - This is a great way to start; all your code is saved for later
> - You will be able to run the file you create from within RStudio or using R’s source() function."

implying that the best way to work in R is to start in a .R file (which I agree with) rather than directly in the console.

However, in lesson 2, the author wrote 

> "The most effective way I find to work in R, is to play around in the interactive session, then copy commands across to a script file when I’m sure they work and do what I want"

implying that the best way to work in R is to work directly in the console rather than starting in a .R file and running commands from there.

These two pieces of advice are inconsistent, and I'd suggest editing the second piece of advice to reflect the first.

naupaka:
Yes, I completely agree. A pull request to change the second to match the first would be most welcome. I think too much time at the console instead of in a script leads to bad habits.

rlbarter:
Happy to submit a pull request. First a quick question from a contributing newbie: when making changes, should I edit the .Rmd file in `_episodes_rmd/` or the .md file in `_episodes/` or both (or neither)?. Do I need to recompile anything before submitting the PR? 

I couldn't find the answers to these questions in the contributing.md file - perhaps it is there and I just missed it?

naupaka:
Hi @rlbarter just edit the `Rmd` file(s). And no need to recompile or regenerate anything. I do that every few weeks after a few PRs have accumulated. I also went looking for the details on how to contribute to the R lessons (which are slightly different than all other lessons because of the use of `Rmd`), and I can't find it either. I recall there used to be helpful R-specific guidelines in there somewhere, so perhaps someone has changed something that I'm not aware of...

In any event - just modify, add, and commit the `Rmd` files you'd like to change and then I'll take a look over them and we'll go from there!

naupaka:
I opened a new issue #304 about the issue with the contributing.md guidelines. Thanks for pointing that out.

PR #305, opened 2017-08-21
Updated advice about writing code in script rather than console

rlbarter:
This pull request is to update the advice in episode 2 about running R code from a script rather than a console. It is in reference to issue [#301](https://github.com/swcarpentry/r-novice-gapminder/issues/301).


naupaka:
Thanks! Much appreciated. Closes #301

PR #306, opened 2017-08-23
Ensure typographic consistency + reformulate sentences

shippy:
Typographic changes made the following things consistent:
* spaces after commas in function calls
* backticks when referring to packages, functions, and variable/column names
* numerals expressed in words when they are less than 10
* capitalization convention of "observation variables"

I've also rewritten some sentences/paragrahs to make the sentence logic clearer.

(Since some of the edits required reflowing the paragraphs, I recommend viewing this with `git diff --word-diff`.)


naupaka:
Excellent. Thanks for these improvements!

PR #310, opened 2017-09-25
Fixed bug in challenge #1 solution

granek:
In 11-writing-data.Rmd the proposed solution to challenge #1 used "p" without first assigning to it. This fixes that bug.

### DIFF
--- a/_episodes_rmd/11-writing-data.Rmd
+++ b/_episodes_rmd/11-writing-data.Rmd
@@ -71,7 +71,7 @@ Open up this document and have a look.
 > > ```{r, eval = FALSE}
 > > pdf("Life_Exp_vs_time.pdf", width = 12, height = 4)
 > > 
-> > ggplot(data = gapminder, aes(x = year, y = lifeExp, colour = country)) + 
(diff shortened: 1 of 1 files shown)


_episodes_rmd/11-writing-data.Rmd
```diff
@@ -71,7 +71,7 @@ Open up this document and have a look.
 > > ```{r, eval = FALSE}
 > > pdf("Life_Exp_vs_time.pdf", width = 12, height = 4)
 > > 
-> > ggplot(data = gapminder, aes(x = year, y = lifeExp, colour = country)) + 
+> > p = ggplot(data = gapminder, aes(x = year, y = lifeExp, colour = country)) +
```
naupaka: We try to use the assignment operator `<-` instead of `=` outside of parameter setting. Can you change it here?

naupaka:
Thanks! Good catch. We try to follow the [Google R Style guide](https://google.github.io/styleguide/Rguide.xml), which suggests the use of the assignment operator `<-`. If you can change that, I'll merge.

naupaka:
Thanks!

PR #316, opened 2017-10-18
formatting a list of things as a list

weaverbel:
I think this improves readability


tomwright01:
I agree it improves readabilty, however it doesn't differentiate between what the different options actually do (the original wording doesn't either). I think it would be improved if it could read something like:

1. Click the `Run` button ...  to run the single current line.
2. Select `Run lines` ... to run all selected lines of code.
3. Press `Ctrl + enter` ... to run all code in a section.

weaverbel:
I like those suggestions @tomwright01 - you are welcome to add them to my PR.

mawds:
Thanks for the PR.  I've resolved the conflict (which was due to the new `<kbd>` formatting)

PR #317, opened 2017-10-21
print plot and remove buggy formatting

### LINKED ISSUE #314: ep11 (writing data) - format & content issue in challenge 1
In the "Saving Plots" exercise, Challenge 1, there is a formatting error that is making "> >" appear where the authors intend a blank line inside a block quote, and in that same block, the instructions say to "create a second page" but the solution only outputs one page (p is assigned and a modified version is printed, but p itself is not printed).

mfoos:
fixes #314 

I looked into Yihui's suggestion that some whitespace could work until the next stable release, but there are a couple layers of rendering/styling that I couldn't easily replicate so I just took out the empty lines for now


naupaka:
Thanks!

PR #323, opened 2017-11-03
instructor checkout challenge added 05-data-structures-part2.Rmd

AntonioJBT:
Hello!
I'm sending a PR for a new challenge adding the use of tail() and how to get a few random rows from a data frame in the episode 05-data-structures-part2.Rmd.
The main reason behind it is to get students to think a little bit more when checking data frames and not settling with the first few lines only. I think it is better practice, but may add to an already full session. 
Happy to hear any thoughts even if the changes are not accepted.
Thanks in advance,
Antonio

### DIFF
--- a/_episodes_rmd/05-data-structures-part2.Rmd
+++ b/_episodes_rmd/05-data-structures-part2.Rmd
@@ -1,27 +1,25 @@
 ---
 title: "Exploring Data Frames"
-keypoints:
-- Use `cbind()` to add a new column to a data frame.
-- Use `rbind()` to add a new row to a data frame.
-- Remove rows from a data frame.
-- Use `na.omit()` to remove rows from a data frame with `NA` values.
-- Use `levels()` and `as.character()` to explore and manipulate factors
-- Use `str()`, `nrow()`, `ncol()`, `dim()`, `colnames()`, `rownames()`, `head()` and
-  `typeof()` to understand structure of the data frame
-- Read in a csv file using `read.csv()`
-- Understand `length()` of a data frame
-objectives:
-- Be able to add and remove rows and columns.
-- Be able to remove rows with `NA` values.
-- Be able to append two data frames
-- Be able to articulate what a `factor` is and how to convert between `factor` and
-  `character`.
-- Be able to find basic properties of a data frames including size, class or type
-  of the columns, names, and first few rows.
-questions: How can I manipulate a data frame?
-source: Rmd
 teaching: 20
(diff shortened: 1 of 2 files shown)


_episodes_rmd/05-data-structures-part2.Rmd
```diff
@@ -292,10 +294,35 @@ to start digging into our data proper. Check out the first few lines:
...
+> > ~~~
+> > tail(gapminder)
+> > tail(gapminder, n = 15)
+> > ~~~
+> > {: .r}
+> > 
+> > What about a few arbitrary rows just for sanity (or insanity depending on your view)?
+> > 
+> > ~~~
+> > gapminder[sample(nrow(gapminder), 5), ]
```
naupaka: Perhaps add a little more explanation to this syntax -- it might be a little much to parse for novices.

_episodes_rmd/05-data-structures-part2.Rmd
```diff
@@ -1,25 +1,27 @@
...
+- Be able to remove rows with `NA` values.
+- Be able to append two data frames
+- Be able to articulate what a `factor` is and how to convert between `factor` and
+  `character`.
+- Be able to find basic properties of a data frames including size, class or type
+  of the columns, names, and first few rows.
+questions: How can I manipulate a data frame?
 source: Rmd
+teaching: 20
+exercises: 10
```
naupaka: I am not sure why you reordered/changed the YAML header here. Removing the quotes from around each of the list elements will cause problems for the parser - can you change this part back to the way it was originally?

AntonioJBT: Hi @naupaka , 
I didn't, at least not intentionally. I don't really work with Rmd files. I think my Rstudio might have made some changes, I'll check the default setup I have. Clearly I changed this somehow though. Apologies!

_episodes_rmd/05-data-structures-part2.Rmd
AntonioJBT: Thanks, will do.

AntonioJBT:
Hi @naupaka 
Sent a new commit. Let me know what you think:
- The YAML header should be back to what it was.
- I added an explanation to sample(). Is this satisfactory/correct/appropriate?
- I assume the md file will get overwritten. Let me know if I should undo changes though.
Thanks,
Antonio

AntonioJBT:
Thanks!

PR #326, opened 2017-11-28
Fix formatting

ErinBecker:
There were a couple of challenges that weren't rendering properly and one that was misnumbered. This PR fixes those issues.


naupaka:
Excellent. Thanks!

PR #327, opened 2017-11-28
12 plyr patch 1

ErinBecker:
Moves challenge 1 from end of episode to immediately after first relevant example. This gives learners a chance to practice the code structure they have just seen before all of the nuances are introduced.


naupaka:
Thanks!

Issue #307, opened 2017-08-29
Missing instructions for where the cats data frame comes from


sandrabrosda:
I was helping with this R session and there was confusion where to get the cats data frame from or how to create it.
Maybe it should be explained first how to create a data frame from scratch and then go into how to manipulate it.
The missing command for the cats data frame could be
cats <- data.frame(coat = c("calico", "black", "tabby"), weight = c(2.1, 5.0,3.2), likes_string = c(1, 0, 1))

Sandra

naupaka:
Hi @sandrabrosda thanks for the note. The beginning of [episode 4](http://swcarpentry.github.io/r-novice-gapminder/04-data-structures-part1/) has instructions for making this file in a text editor. Presumably this would go along with some discussion of the csv format. Creating the data frame directly in R would then obviate the need for the bit on `read.csv()`. 

However, I agree it is a bit odd to ask folks to create a csv by hand, since this this something they will almost never do in normal use. I think it might be reasonable to change this lesson to have the data frame created within R instead, with the code you included. Would you be willing to submit a PR for that change that I could review?

sandrabrosda:
Hi Naupaka,

I’m happy to submit a PR for that change.
However, I have to figure out how to do it first and I’m away for almost the rest of the month. But if you’re happy for me to do it next month I’ll do it.

Regards,
Sandra

naupaka:
Sure, no rush. There are some instructions for contributing, [here](https://github.com/swcarpentry/r-novice-gapminder/blob/gh-pages/CONTRIBUTING.md). I'm also happy to help once you get to that point.

missaugustina:
One way I did it for the instructor training course I took was using "download.file" in an R notebook. 
`
download.file(
  "https://raw.githubusercontent.com/swcarpentry/r-novice-gapminder/gh-pages/_episodes_rmd/data/feline-data.csv", "cats_swcarp.csv")

cats <- read.csv("cats_swcarp.csv")
`

naupaka:
@missaugustina you can also read csv directly from a URL if you like.

```
cats <- read.csv("https://raw.githubusercontent.com/swcarpentry/r-novice-gapminder/gh-pages/_episodes_rmd/data/feline-data.csv")
```

missaugustina:
I don't recommend reading directly from the url for 2 reasons: 1) possible wifi issues and 2) a local copy ensures you know what version of the file you are working with.

mfoos:
[...]

PR #335, opened 2018-01-15
Update 04-data-structures-part1.md

sandrabrosda:
Updated the lesson as discussed in issue #307 to include a way to create the cats data frame in R and save it as a csv file in the data folder.

### DIFF
--- a/_episodes/04-data-structures-part1.md
+++ b/_episodes/04-data-structures-part1.md
@@ -23,6 +23,13 @@ One of R's most powerful features is its ability to deal with tabular data -
 such as you may already have in a spreadsheet or a CSV file. Let's start by
 making a toy dataset in your `data/` directory, called `feline-data.csv`:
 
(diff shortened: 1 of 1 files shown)


naupaka:
@sandrabrosda Thanks for the PR! I like the idea of writing out the csv after creating it. 

Two things: 

1) I think you might want to add `row.names = FALSE` as an extra parameter to `write.csv()`, because otherwise the first column will be the row numbers, and 
2) it looks like you've modified the `md` instead of the `Rmd`. The md files get regenerated from the `Rmd` when we rebuild the site, so any permanent changes need to be made to (just) the `Rmd`. 

Can you update the PR to fix these?

sandrabrosda:
Ok I think I did that now. Sorry for that. GIT still confuses me.

naupaka:
Revised PR in #338

PR #336, opened 2018-01-15
name of dir needs an s as the instruction

orchid00:
should be scripts dir

### DIFF
--- a/_episodes/05-data-structures-part2.md
+++ b/_episodes/05-data-structures-part2.md
@@ -673,7 +673,7 @@ into a script file so we can come back to it later.
 > as its argument (or by pressing the "source" button in RStudio).
 >
 > > ## Solution to Challenge 4
-> > The contents of `script/load-gapminder.R`:
+> > The contents of `scripts/load-gapminder.R`:
 > > 
 > > ~~~
 > > download.file("https://raw.githubusercontent.com/swcarpentry/r-novice-gapminder/gh-pages/_episodes_rmd/data/gapminder-FiveYearData.csv", destfile = "data/gapminder-FiveYearData.csv")


naupaka:
@orchid00 thanks for the PR! That indeed needs to be fixed. However, it looks like you made a change to the markdown files instead of the `Rmd`. Can you update the PR to modify the Rmd instead?

orchid00:
Done, 

@naupaka if you have a few min, could you please shortly explain me why do we have md if the Rmd is the one rendered for the site?

naupaka:
So we use a (GitHub provided through [GitHub pages](https://pages.github.com/)) static website renderer called [Jekyll](https://jekyllrb.com/) that turns the markdown files into html, which is what people see when they go to the rendered lessons. The problem is that Jekyll can't parse Rmd files, so those have to be converted locally (we use `rmarkdown::render()`) by the maintainers using the Makefile in the repository. So I or one of the other maintainers pull down all the new changes that people have contributed, then convert them from Rmd to markdown, then push that back to GitHub, which then is set up to render the markdown files into the final html web pages.  This process overwrites any existing markdown, which is why it's important for the changes to happen to the Rmd files. As for why Rmd instead of md, it's because it ensures that the code and the output of the code don't get out of sync or have to be copied into the output sections since the output is generated from the code directly when it is rendered by the rmarkdown package.

Hope that helps clarify a bit? We also have some info on how that all happens for the Carpentry lesson [in this lesson example module](http://swcarpentry.github.io/lesson-example/).

naupaka:
Closed per #337

orchid00:
Thank you @naupaka!

PR #338, opened 2018-01-17

sandrabrosda:
Updated the lesson as discussed in issue #307 and PR #335 to include a way to create the cats data frame in R and save it as a csv file in the data folder.


naupaka:
Thanks! Revision to #335. Closes #307.

PR #340, opened 2018-01-18
Add a file pointing contributors to the correct location to make lesson changes

mawds:
This PR adds a file to episodes/ which points contributors to episodes_rmd/ as the correct place to make changes in the R lesson.

### DIFF
--- a/_episodes/000_DO_NOT_EDIT_FILES_IN_THIS_DIRECTORY_README_FOR_MORE_DETAILS.txt
+++ b/_episodes/000_DO_NOT_EDIT_FILES_IN_THIS_DIRECTORY_README_FOR_MORE_DETAILS.txt
@@ -1,0 +1,12 @@
+If you are making a change to this lesson, please edit the corresponding
+.Rmd file in _episodes_rmd/.
+
+The .md files in this directory are created automatically as part of 
+the lesson build process, so any changes made here will be overwritten.
+
+The episode 5 of the lesson example:
+http://swcarpentry.github.io/lesson-example/ contains details of
+how to include RMarkdown in an episode.
+
+Thanks in advance for your contribution!
+


naupaka:
This seems like a useful thing, but I think it has the potential to cause problems with the Jekyll build process. 

@rgaiacs is this going to turn into a rendered page or mess things up in other ways?

rgaiacs:
@mawds Thanks for the pull request.

As @naupaka mentioned, https://github.com/swcarpentry/styles/blob/gh-pages/_includes/navbar.html#L72 has a bug. :-( Or one of the files is missing some information on the YAML header.

> is this going to turn into a rendered page or mess things up in other ways?

I can't say by head. :-(

mawds:
As far as I can tell it doesn't break anything - I've merged the PR into my local copy of r-novice-gapminder at https://mawds.github.io/r-novice-gapminder/ and everything seems to work OK.  You don't get a rendered version of it in the episode list

@rgaiacs  - what's the bug in navbar.html?

rgaiacs:
Hi @mawds, I didn't understand exactly the problem that you were trying to solve.

> points contributors to episodes_rmd/ as the correct place to make changes in the R lesson.

This is similar to a problem that I have. Users will submit pull request changing `_includes`, `_layouts`, `bin` to the lesson instead of https://github.com/swcarpentry/styles.

The solution to my problem, based with some discussion with Jonah Duckles last year, is to use submodules despite people hating it.

The solution to your problem is to remove `_episodes` and have it built by a continuous integration server. This is one of the reasons that I wish that we move to GitLab.

For the moment, what I want to do is https://github.com/swcarpentry/styles/issues/198. For `_episodes`, would be possible to change https://github.com/swcarpentry/r-novice-gapminder/blob/gh-pages/bin/generate_md_episodes.R and include a comment as the first thing in the YAML header (see http://www.yaml.org/spec/1.2/spec.html#id2780069).

mawds:
The problem is that some PRs make changes to `_episodes/xx-lesson.md` rather than `_episodes_rmd/xx-lesson.Rmd`.  This PR was just to put a file in `_episodes` reminding people not to edit the files there, but in `_episodes_rmd`.

The better answer is, as you say, to use CI so there's no temptation to edit  `_episodes`.  Until that happens this seemed like a simple way of nudging people to the right location, provided it doesn't break anything in the existing build process.

>  For _episodes, would be possible to change https://github.com/swcarpentry/r-novice-gapminder/blob/gh-pages/bin/generate_md_episodes.R and include a comment as the first thing in the YAML header (see http://www.yaml.org/spec/1.2/spec.html#id2780069).

Yes - I think that would work too (or in addition to this approach).

naupaka:
I'm ok with merging this if you think it won't break anything, @rgaiacs 

I also think the comment in the YAML would be good as well. This is by far the most common mistake people make when contributing to these lessons.

rgaiacs:
> I'm ok with merging this if you think it won't break anything

No reasons to break anything.

naupaka:
OK I am going to merge this. If there are any issues, it's easy enough to revert. Thanks for the contribution, @mawds!

PR #341, opened 2018-01-25
Use kbd formatting in episode 01

mawds:
This updates episode 01 to use the HTML `<kbd>`  tag.


ErinBecker:
Thanks @mawds and @jcoliver !

Issue #104, opened 2016-05-06
one day approach to teaching material


pschloss:
I just finished teaching a one day version (~6 hrs including breaks) of these materials. I thought it might be useful to share what I did for others that might be looking for a trimmed down version of these thorough materials. The workshop started with a morning of bash, then an afternoon and morning of R, and then an afternoon of git. That turned out to be a mistake for the person teaching git, but whatever. My motivation was to give people the minimum that they needed to get going with R.
- Lesson 1: I quickly went over the material without using git
- Lesson 2: skipped
- Lesson 4: More or less went up to but not including factors
- Lesson 5: Combined with the material from Lesson 4
- Lesson 9: Combined with the material from Lesson 4
- Lesson 3: Discussed using the help window in RStudio to find help with `read.csv`
- Lesson 6: Skipped using `which`, "Handling special values", "Factor subsetting", "Matrix subsetting", "List subsetting"
- Lesson 7: Got to the gdp calculator at the end of the first day and picked up there on the second day
- Lesson 10: Only taught `if ... else` in the context of the more sophisticated gdp calculator
- Lesson 11: skipped
- Lesson 12: skipped
- Lesson 13: Followed pretty literally
- Lesson 8: Followed pretty literally
- Lesson 14: skipped
- Lesson 15: skipped
- I also added about 20-30 min on `knitr` showing how the plots from Lesson 8 could be put into a document.

I realize that I skipped lists and matrices and barely introduced factors to say they are categorical data types. To get through the `dplyr` and `ggplot2` stuff those just aren't needed and you can go a long way in R without needing them.

When teaching the function component I tried to build up the gdp calculator piece by piece. I would show them how to do the year and have them do the country. Towards the end of the first day I could sense that they weren't getting it and they were glazing over. So I had them get in pairs and alternate explaining each line to each other. They really perked up and seemed to have more confidence. We repeated this the following morning to rebuild what we had done and to go forward with the `if` statements.

Having taught some version of these materials twice now, I fear that a lot of the SWC materials have become bloated beyond what is truly necessary to get someone going and so that kind of effected what and how I approached the materials.

naupaka:
Awesome, thanks for the notes @pschloss! 

Any chance you have some of that `knitr` content written up and would care to share it via PR? Issue #11 is about adding a `knitr` lesson - @kbroman made a set a while back and we still haven't brought them in although in my experience many people teach `knitr` at least briefly in R workshops even without a formal lesson.

[...]

PR #345, opened 2018-02-01
Include outline lesson plans

mawds:
This includes some of the paths through the material  that were suggested in issue #104 in the instructor guide


jcoliver:
Looks good to me. Any objections, @naupaka?
After the pull, we could also add a _very_ brief bit of information for each path about intended audience(s).

naupaka:
Looks great. +1 to merge

PR #354, opened 2018-03-08
fix objectives 08-plot-ggplot2

sashasaurus:
Changed original objective (To understand the basic grammar of graphics, including the aesthetics and geometry layers, adding statistics, transforming scales, and coloring or panelling by groups") to more measurable verbs: apply, manipulate, improve


mawds:
Thanks for the PR; that makes the objectives much clearer

PR #358, opened 2018-03-18
fix one erratum

prosoitos:
factors *are* vectors (can be verified by testing a factor x with `is.atomic(x) || is.list(x)` as per http://adv-r.had.co.nz/Data-structures.html)

### DIFF
--- a/_episodes_rmd/05-data-structures-part2.Rmd
+++ b/_episodes_rmd/05-data-structures-part2.Rmd
@@ -164,7 +164,7 @@ cats <- na.omit(cats)
 ## Appending to a data frame
 
 The key to remember when adding data to a data frame is that *columns are
-vectors or factors, and rows are lists.* We can also glue two data frames
+vectors and rows are lists.* We can also glue two data frames
 together with `rbind`:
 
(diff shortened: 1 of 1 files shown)


jcoliver:
Thanks, @prosoitos . This sounds good to me - any objections @mawds @naupaka @tomwright01 ?

prosoitos:
To add more info, according to http://127.0.0.1:25436/library/base/html/factor.html, "factor" is the "class" of the object (as "character" or "double"). And according to http://r4ds.had.co.nz/vectors.html, "factors" are "augmented vectors built on top of integer vectors".

naupaka:
I am 100% in support of this change.

prosoitos:
Great. Thank you all! I am in the middle of my Carpentry instructor training and I spotted other little issues in this lesson. So I will send another PR with other little suggestions later. Apologies for not doing them all at once!

naupaka:
@prosoitos No worries! Actually we prefer them this way - it keeps the PRs focused and easy to review.

prosoitos:
Ok. Good to know.

PR #362, opened 2018-03-26
Add section demonstrating how to remove columns from a df

JamieMcDevittIrwin:

---

### DIFF
--- a/_episodes_rmd/05-data-structures-part2.Rmd
+++ b/_episodes_rmd/05-data-structures-part2.Rmd
@@ -159,6 +159,23 @@ Let's reassign the output to `cats`, so that our changes will be permanent:
 cats <- na.omit(cats)
 ```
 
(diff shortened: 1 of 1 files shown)


jcoliver:
Thanks, @JamieMcDevittIrwin ! I _think_ the second example needs a comma:
<pre>
-cats[!drop]
+cats[, !drop]
</pre>
Right?

JamieMcDevittIrwin:
When I run the code, it works either way (with or without the comma) and the output is still a data frame. But I think you're right, it might be best for consistency to keep the comma!

jcoliver:
Thanks, @JamieMcDevittIrwin . The single-index subsetting of data frames gets introduced at the very end of the next lesson (06-data-subsetting). I was figuring to keep on with the `[ , ]` syntax in this part of the lesson.

PR #364, opened 2018-04-01
Use the tidyverse

maurolepore:
It looks like the lessons are gradually shifting to the tidyverse. This PR replaces -- as much as possible -- functions from base R with functions from the tidyverse.


naupaka:
Thanks for the suggestion, @maurolepore. I think that a change of this scale would need to be discussed and agreed upon more broadly before we made such a shift. The general approach has been to use base R in all cases except where a lesson explicitly focuses on a package like `dplyr`, `tidyr`, or `ggplot2`. 

There are strong feelings on both sides of this debate (lead with base or lead with tidy), and I think any such philosophical shift towards all tidyverse all the time would need to be agreed upon more broadly by the community of instructors first, in which case it might make more sense to move this over to an issue for further discussion. 

Also note that PRs should only include the changes to the Rmd files; there's no need to also include your `.Rhistory`, `.Rproj`, or `.gitignore`.

What do you think @jcoliver @mawds?

maurolepore:
Thanks @naupaka for your comments. I agree that this a comlete shift to the tidyverse is big deal and deserves more discussion (https://github.com/datacarpentry/R-ecology-lesson/issues/378). 

> The general approach has been to use base R in all cases except where a lesson explicitly focuses on a package like dplyr, tidyr, or ggplot2.

I think that's a good approach. An additional criteria to decide when to teach a tidyverse function might be when that function solves a common problem of base R. 

For example, the difference between `base::rbind()` and `tibble::add_row()` isn't important. But it is important the difference between `utils::read.csv()` and `readr::read_csv()` because `utils::read.csv()` causes many problems. 

In my opinion, the key behaviours that determine the difference between frustration and success are relatively few (e.g. remember to set `stringsAsFactors = FALSE`). And we can shape the path to success by providing the learners with an environment that enforces the key behaviour (e.g. promoting the use of `readr::read_csv()`, which never converts strings to factors).

> Also note that PRs should only include the changes to the Rmd files; there's no need to also include your `.Rhistory`, `.Rproj`, or `.gitignore`.

Ups, sorry!

jcoliver:
Closing pending discussion at #365

PR #371, opened 2018-04-13
Update 07-control-flow.md

raynamharris:
removed comments about "Poisson distribution" and "lamda = 8" that we left over when 
`x <- rpois(1, lambda=8)` was changed to `x <- 8` in 
https://github.com/swcarpentry/r-novice-gapminder/commit/f60708b030f3c4479536476c3647080adfee129c


raynamharris:
Eh, I just realized I changed the .md file and not the .Rmd file. I'll fix it.

PR #372, opened 2018-04-13
remove out of date comment in 07-control-flow.Rmd

raynamharris:
removed comments about "Poisson distribution" and "lamda = 8" that we left over when 
`x <- rpois(1, lambda=8)` was changed to `x <- 8` in 
https://github.com/swcarpentry/r-novice-gapminder/commit/f60708b030f3c4479536476c3647080adfee129c


naupaka:
Thanks!

Issue #366, opened 2018-04-10
rename episodes


raynamharris:
I really like the title of lessons 13-15 because they contain the name of the package and what it does. 

I think it is worth renaming the `ggplot` and `dplyr` lessons (eps. 8 & 12) for 3 reasons. 
1. This will make it more consistent with the file name for the episode, 
1. it will help users find what they need, and 
1. the questions column will be a better place to describe in more detail why this tool is useful. 

Is anyone else in favor of renaming the title of these episodes?

![screen shot 2018-04-10 at 12 32 40 am](https://user-images.githubusercontent.com/7544197/38534901-26bb13f6-3c57-11e8-9241-2e11c423c5b9.png)

jcoliver:
Thanks, @raynamharris ! So what might we use as replacements? 

8. Publication-quality Graphics with ggplot
12. Split-Apply-Combine with dplyr

Although I'm not sold on my suggestion for 12...

naupaka:
I am definitely in favor of renaming to include the package name - will help with SEO too. I like @raynamharris's suggestion for the second one: "Splitting and Combining Data Frames with dplyr"

raynamharris:
Thanks @naupaka. I went with my suggestion for dplyr and @jcoliver 's suggestion for ggplot2 in PR #374

naupaka:
Looks like I was wrong! It's a plyr lesson not dplyr. Sorry for the confusion.

raynamharris:
Ah. Your right! I corrected the typo.

naupaka:
Closed by #374

PR #374, opened 2018-04-15
rename ggplot2 and plyr episodes

raynamharris:
renames the ggplot2 and plyr episodes in the Rmd files and the guide as per suggestions in issue #366

### DIFF
--- a/_episodes_rmd/12-plyr.Rmd
+++ b/_episodes_rmd/12-plyr.Rmd
@@ -1,5 +1,5 @@
 ---
-title: Splitting and Combining Data Frames with dplyr
+title: Splitting and Combining Data Frames with plyr
 teaching: 30
 exercises: 30
 questions:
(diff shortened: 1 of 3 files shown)


_episodes_rmd/12-plyr.Rmd
```diff
@@ -1,5 +1,5 @@
 ---
-title: Split-Apply-Combine
+title: Splitting and Combining Data Frames with dplyr
```
naupaka: So it looks like I was wrong. This is a `plyr` lesson, not `dplyr`. Can you make that adjustment?

raynamharris: Good catch! These too names are too similar!

naupaka:
Thanks!

naupaka:
Closes #366

PR #387, opened 2018-05-27
Tip: Running examples with ctrl+enter

MilesMcBain:
I wasn't sure how to make a tip list, so it could be that this can be pulled into a list with the other tip on R functions.


naupaka:
Hi @MilesMcBain thanks for the PR. I might swap the order, so the bit on running examples comes after 'reading help' instead of the other way around. That way it's more in line with how the help itself is presented.

PR #389, opened 2018-05-27
Make it clear csv output is from feline-data.csv

MilesMcBain:
The output shown after the toy dataset creation is not shown to the user unless they view the `feline-data.csv` file. This text tries to makes that clear.


naupaka:
Thanks!

PR #390, opened 2018-05-27
Strip back explanation of read.* functions.

MilesMcBain:
The explanation here is potentially confusing as it begins as if the leaner had already encountered `read.table`. It uses many words to rapidly introduce many concepts related to reading data. None of these are relevant to the rest of the lesson. Perhaps it would be better just to make the simple point that there are alternatives to read.csv.

### DIFF
--- a/_episodes_rmd/04-data-structures-part1.Rmd
+++ b/_episodes_rmd/04-data-structures-part1.Rmd
@@ -62,16 +62,9 @@ cats <- read.csv(file = "data/feline-data.csv")
 cats
 ```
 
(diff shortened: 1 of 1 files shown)


naupaka:
@MilesMcBain I actually disagree with you on this one. I think it is useful to have that extra information in the lesson. I think if instructors want to skip over it while teaching, then that's fine, but I think it is helpful and useful info for many workshops as well as for people using the lessons to self-teach.

jcoliver:
Closing for now, feel free to resubmit following tidyverse discussion.

PR #391, opened 2018-05-27
Remove reference to `[]` being a function and tweak.

MilesMcBain:
Yes it is a function, but this seems like an awkward spot for a sidebar on how all operators are in fact functions, at this point, learners probably can't appreciate that. So it's just a distraction.

It also seems contradictory to say `[]` means "get me the nth element" when the very next step is to use it to get multiple elements. "get me these elements" works for a single element and also works for logical indexing.

### DIFF
--- a/_episodes_rmd/06-data-subsetting.Rmd
+++ b/_episodes_rmd/06-data-subsetting.Rmd
@@ -59,10 +59,8 @@ x[1]
 x[4]
 ```
 
-It may look different, but the square brackets operator is a function. For vectors
-(and matrices), it means "get me the nth element".
-
-We can ask for multiple elements at once:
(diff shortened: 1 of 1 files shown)


naupaka:
I agree with the second part of the proposed change (from "get me the nth" to get me these") but I am not so sure it is worth removing the statement about `[` being a function. Again, an instructor is free to skip over this when teaching, but the additional info could be useful, and helpful for learning about how R works as a language (which might be important for people who have programming experience in another language and are attending a workshop or using these materials to learn R).

naupaka:
Closing for now in order to prepare for Travis changeover, but feel free to re-open after making the suggested modifications.

PR #393, opened 2018-05-29
Update 14-tidyr.Rmd

MilesMcBain:
Wide format aka 'tidy data' is the standard format for modelling and that is probably worth mentioning. Long format has uses with ggplot2.  

Readability is probably a concern but secondary to these two.


MilesMcBain:
Actually this is probably not a great PR since the rest of the lesson uses this more 'wide' version. I guess I am talking about your 'intermediate' form with respect to modelling.

Issue #399, opened 2018-06-29
Gapminder 5-year data missing


maglet:
I tried to load the gapminder data from the URL in the tutorial (https://swcarpentry.github.io/r-novice-gapminder/data/gapminder-FiveYearData.csv) and i got an error saying the file was not found. 

Looking in the data folder in the repo, it appears that only the wide data table is there.  Can someone add the long data back?

- Tobin

jcoliver:
Thanks, Tobin @maglet ! Working on it #400 . In the meantime, you can get the gapminder data from: https://raw.githubusercontent.com/resbaz/r-novice-gapminder-files/master/data/gapminder-FiveYearData.csv

PR #400, opened 2018-06-29
Add long-ish gapminder data

jcoliver:
Add the long-ish form of gapminder data, per #399 . An alternative to this PR would be to change the URL referenced in episode 5 ~The Empire Str~ Exploring Data Frames to the source where I downloaded the gapminder data: https://raw.githubusercontent.com/resbaz/r-novice-gapminder-files/master/data/gapminder-FiveYearData.csv. This latter option might be preferable...?


naupaka:
I think it's probably better to have the data in this repo and update lesson links to point to it here. That link (resbaz) is from the original source of the lessons that we forked to make the SWC version years ago.

PR #408, opened 2018-07-13
Update 06-data-subsetting.Rmd

vickenhillis:
Deletes the solution to challenge 1 that involves subsetting by name, because subsetting by name is not introduced until the subsequent section.

### DIFF
--- a/_episodes_rmd/06-data-subsetting.Rmd
+++ b/_episodes_rmd/06-data-subsetting.Rmd
@@ -168,13 +168,13 @@ x
 > print(x)
 > ```
 >
-> Come up with at least 3 different commands that will produce the following output:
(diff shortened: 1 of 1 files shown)


vickenhillis:
May be confusing to learners to present a solution to a challenge that involves subsetting by name when that concept is not introduced until the subsequent section.

Alternatively, I guess one might argue that using a solution that involves a coding concept that hasn't yet been introduced provides some useful foregrounding and repetition. 

But, my guess is it would generate confusion.

naupaka:
Thanks for the suggestion. I could go either way on this. @jcoliver @mawds ?

naupaka:
OK, sounds good to me. Thanks @vickenhillis!

PR #416, opened 2018-08-22
update gapminder_data.csv path

orchid00:
link to this https://github.com/swcarpentry/r-novice-gapminder/issues/410
@naupaka

PR #417, opened 2018-08-23
update location of the file gapminder_data.csv

orchid00:
Hi @jcoliver @naupaka 
after previous PR https://github.com/swcarpentry/r-novice-gapminder/pull/416, I just noticed that the file gapminder csv file is in two different locations
Is that needed?

https://github.com/swcarpentry/r-novice-gapminder/tree/gh-pages/_episodes_rmd/data
https://github.com/swcarpentry/r-novice-gapminder/tree/gh-pages/data

Do we also need to change the name on the _episodes_rms/data folder?

### DIFF
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -83,7 +83,7 @@ git pull data master
 {: .bash}
 
 If Git is not being taught as part of the workshop the raw data can be downloaded from
-[gapminder-FiveYearData][gapminder-data] and
+[gapminder_data][gapminder-data] and
 [gapminder-FiveYearData-Wide][gapminder-data-wide].
 
 Attendees can use the `File - Save As` dialog in their browser to save the file.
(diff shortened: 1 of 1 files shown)


orchid00:
in case it is needed the PR is here for _episodes_rmd/data 
https://github.com/swcarpentry/r-novice-gapminder/pull/418

jcoliver:
Thanks for the PR, @orchid00 . Before I accept, I want to check with @naupaka to make sure the files in `episodes_rmd/data` aren't auto-generated.

naupaka:
@jcoliver nothing in `_episodes_rmd/` shoud be autogenerated; everything in `_episodes/` is or should be. I think `_episodes_rmd/data` is the preferred place for the files and we should probably get rid of `data/`, but I am not 100% sure on that. Since knitting happens relative to the path of the Rmd files, I think the two files in `data/` never get used. I think we should probably merge these PRs, then delete the `data/` folder, re-run the makefile to check if it all builds. If not, then we can always quickly revert...

fmichonneau:
the appropriate location for the data files is `_episodes_rmd/data`

jcoliver:
We'll need to check other episodes (see #421 ) to make sure the file name change is accommodated there, too.

naupaka:
This is superseded by #424 so I am going to close. Thanks for the help, @orchid00 @jcoliver @fmichonneau

PR #418, opened 2018-08-28
Rename gapminder-FiveYearData.csv to gapminder_data.csv

orchid00:


naupaka:
Merging. Will fix rest of paths soon. Not merging #417 because that still uses incorrect path `data/` vs `_episodes_rmd/data`

PR #422, opened 2018-08-29
don't evaluate rm() during rendering

fmichonneau:
This is causing issues with recent upstream changes to the script building the lessons. Running `rm()` removes variables that are needed by the script.


jcoliver:
Since the results of calling this function are never investigated, skipping evaluation should not have downstream effects.

PR #441, opened 2018-10-15
Update 08-plot-ggplot2.Rmd

markdunning:
Fix some inconsistencies about the way in which the package name ggplot2 is formatted

### DIFF
--- a/_episodes_rmd/08-plot-ggplot2.Rmd
+++ b/_episodes_rmd/08-plot-ggplot2.Rmd
@@ -199,7 +199,7 @@ lines.
 
 ## Transformations and statistics
 
-Ggplot also makes it easy to overlay statistical models over the data. To
+ggplot2 also makes it easy to overlay statistical models over the data. To
 demonstrate we'll go back to our first example:
 
 ```{r lifeExp-vs-gdpPercap-scatter3, message=FALSE}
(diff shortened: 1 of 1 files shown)


jcoliver:
Thanks, @markdunning ! We should be consistent when referring to the package name (`ggplot2`), although to be consistent with other episodes (e.g. 12-15), we should consider how we want to format package names. That is, do we use code formatting, i.e. `ggplot2` and `tidyr`, or plain text, ggplot2 and tidyr, when referring to package names. Both are used in a variety of spots in episodes 12-15, but it would be nice to pick one convention and stick to it. @naupaka ? @mawds ?

naupaka:
I say we stick with code styling for package names, like here: http://datacarpentry.org/semester-biology/docs/course/style-guide/

naupaka:
See also the Carpentries Style Guide, although I don't think it speaks to this particular point:
https://docs.google.com/document/d/10KutjajR-5RJfPAF8UQlxzcd8tx4PNFXH0acDFD2dBs/edit

naupaka:
Actually I take that back. We are ostensibly following the tidyverse style guide, which explicitly says to avoid code style for package names:
http://style.tidyverse.org/documentation.html#r-code

PR #446, opened 2018-10-18
mixed up TRUE and FALSE in vector recycling

wrathofquan:


mawds:
Hi.  Good spot; thanks for the PR.

You'll need to make the change in `_episodes_rmd/06-data-subsetting.Rmd` not `_episodes/06-data-subsetting.md` though. The `.Rmd` R Markdown files are used to build the `.md` files when the lesson is built. Please could you modify your PR to make the change in the `.Rmd` file?

PR #452, opened 2018-11-07

Malarkey73:
I think for novices the idea of  "names(cats)" a longer vector being %in%  c("age") a single character vector is extraneous cognitive load. NB the c() is un-necessary too. 
More natural just to ask are the names equal "==". Maybe also change "!drop" to "-drop" as that again is consistent with the previous examples of column selection immediately above.

---

---

### DIFF
--- a/_episodes_rmd/05-data-structures-part2.Rmd
+++ b/_episodes_rmd/05-data-structures-part2.Rmd
@@ -172,8 +172,8 @@ Notice the comma with nothing before it, indicating we want to keep all of the r
 Alternatively, we can drop the column by using the index name.  
 
 ```{r}
-drop <- names(cats) %in% c("age")
-cats[,!drop]
+drop <- names(cats)  == "age"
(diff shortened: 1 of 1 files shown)


naupaka:
@Malarkey73 thanks for the PR! I agree it could be confusing, but the counterargument is that it enables removal of multiple discontiguous columns. The switch from `!drop` to `-drop` is a slightly different issue. Thoughts @jcoliver @mawds ?

jcoliver:
Good point @Malarkey73 . The `%in%` operator is covered more in depth in episode 06, so I agree that using `names(cats) == "age"` would be appropriate at this point in the lesson. As for `-drop` vs. `!drop`...I think keeping the logical operator (`!`) makes sense since the result of `names(cats) == "age"` is a logical vector.

jcoliver:
Closing for now, but feel free to re-open after restricting change to updating only the portion about the `%in%` operator.

PR #454, opened 2018-11-18
Clarify warning vs error of rbind

### LINKED ISSUE #420: Clarify that "invalid factor level: NA generated" is a warning and not an error, and that rbind () is still executed
---
Hi I am a new Instructor trainee and I have been working in the "Working with data frames lesson" (lesson 5). The lesson has the participant use rbind to merge a new row to a dataframe, however, the input for coat variable is not one of the factor levels, which throws a warning and generates an NA value:

newRow <- list("tortoiseshell", 3.3, TRUE, 9)
cats <- rbind(cats, newRow)

Warning in `[<-.factor`(`*tmp*`, ri, value = "tortoiseshell"): invalid
factor level, NA generated 

The lesson then goes on the explain the warning was generated because tortoiseshell was not a pre-defined level in coat, and next shows the participant how to add levels.  However, due to the red text and the lack of explanation that the warning was not an error, it is easy to miss that rbind() was still executed and an NA was put in place of the undefined factor level. I think it would be easy to point this out here, which can be demonstrated by printing cats.  

Instead, the lesson then goes on to explain more about factors, using the str function, and converting factors to characters before "cats" is printed out we the learner sees there is an extra row which they are instructed to remove.  As I worked through this, it was not clear at first where the extra row had come from.  

---

lucymchang:
This pull request is intended to resolve #420 by adding some text reflecting on the results of using `rbind` with factors.  By examining the object (`cats`), the learner can note the unusual results following the warning about factor levels (successful `rbind` but introduction of `NA`), which may better set up the subsequent discussion about what factors are.

### DIFF
--- a/_episodes_rmd/05-data-structures-part2.Rmd
+++ b/_episodes_rmd/05-data-structures-part2.Rmd
@@ -87,7 +87,7 @@ newRow <- list("tortoiseshell", 3.3, TRUE, 9)
 cats <- rbind(cats, newRow)
 ```
 
-Looks like our attempt to `rbind` returns a warning.  Recall that, unlike errors, warnings do not necessarily stop a function from performing its intended action.  You can confirm this by taking a look at the `cats` data frame.
+Looks like our attempt to use the `rbind()` function returns a warning.  Recall that, unlike errors, warnings do not necessarily stop a function from performing its intended action.  You can confirm this by taking a look at the `cats` data frame.
 
 ```{r}
 cats
--- a/_episodes_rmd/05-data-structures-part2.Rmd
+++ b/_episodes_rmd/05-data-structures-part2.Rmd
@@ -87,15 +87,23 @@ newRow <- list("tortoiseshell", 3.3, TRUE, 9)
 cats <- rbind(cats, newRow)
 ```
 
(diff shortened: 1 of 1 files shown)


_episodes_rmd/05-data-structures-part2.Rmd
```diff
@@ -87,15 +87,23 @@ newRow <- list("tortoiseshell", 3.3, TRUE, 9)
 cats <- rbind(cats, newRow)
 ```
 
+Looks like our attempt to `rbind` returns a warning.  Recall that, unlike errors, warnings do not necessarily stop a function from performing its intended action.  You can confirm this by taking a look at the `cats` data frame.
```
naupaka: Usually the style we follow suggests functions are specified with their `()`, so I might suggest changing "attempt to `rbind` returns" to "attempt to use the `rbind()` function returns".

naupaka:
Hi @lucymchang thanks for the PR! If you can make the small style tweak I suggested, I'm happy to merge this.

_episodes_rmd/05-data-structures-part2.Rmd
lucymchang: Hi @naupaka.  Thanks for the comment - I should have been paying more attention to the style guide.  I've added a commit that uses the language you suggested.

naupaka:
Thanks! Closes #420

Issue #430, opened 2018-09-09
Trigonometric functions in radians or degrees


rgaiacs:
The "Mathematical Functions" section starts using the `sin()` function. Trigonometric functions are **not** notice friendly because the arguments could be provided in radians or degrees. I recommend to push the `sin()` example to the end of the section after we introduce the help page so that we could use trigonometric functions as one example for the use of `?`.

## Some examples

```
> sin(0)
[1] 0
> sin(90)
[1] 0.8939967
> sin(1/2)
[1] 0.4794255
> sin(1/2 * pi)
[1] 1
```

## Note

R documentation says

> Angles are in radians, not degrees, for the standard versions (i.e., a right angle is pi/2) ...

jcoliver:
I think the `sin()` function is nice because we can pretty much rely on all audience members to have experience with trig functions outside of R. More so than logarithms and exponents, IMHO. But maybe we could use `sin()` as an example in episode 03 Seeking Help?

cbeleites:
I guess it will depend on the audience whether they work more with trigonometric functions or logarithms. I'd expect biologists to be familiar with exponential growth. Also chemists "meet" logarithms quite regularly, but we hide them (pH = - log_10 c_H⁺, after that, we just use pH.). Nevertheless, trigonometric functions seem to cause less of an "ouch" with many people. 

- It is true that the trigonometric functions working with radian input need explanation (same for log being natural log and not base 10). But IMHO that doesn't necessarily mean they are not suitable for novice courses: I'd think it legitimate to say "I'm not expecting you to actively remember these details - but I hope you'll remember there was something special if you stumble over such behavior".  
I'd embed this in a "standard warning" that many or most programming languages use angles in rad for trigonometric functions and natural logarithm as the most 

- The help episode introduces `c` and `paste` which are IMHO more important than trigonometric or logarithm argument details. 

- If only one of them (sin vs. log) is kept in the introduction, the other would also beautifully fit the paragraph "Know that your code is doing the right thing" in [Writing Good Software](https://swcarpentry.github.io/r-novice-gapminder/16-wrap-up/index.html).  
There, a glimpse into unit testing (i.e. that in addition to typing in and making sure *once* it is possible to write code that checks whether computations are done correctly and that can be run every so often and even automatically) could be given: 

      library (testthat)
      expect_equal (sin (90), 1)

  However, that gets us into a different discussion, namely whether that episode should gain flesh or stay as it is.

[...]

PR #469, opened 2019-02-05
math example arguments for sin and log functions

cbeleites:
- use arguments that more easily challenge possibly erroneous mind models: 
  - sin works on radians, not degrees: easier to recognize with `log (90)` and `log (pi / 2)`
  - log is *not* base 10 log: easy to recognize with `log (10)`

### DIFF
--- a/_episodes_rmd/01-rstudio-intro.Rmd
+++ b/_episodes_rmd/01-rstudio-intro.Rmd
@@ -236,7 +236,7 @@ sin(pi / 2)  # ... work in radians: 90° = π/2 rad
 ```
 
 ```{r}
-log(1)  # natural logarithm
+log(10)  # natural logarithm
 ```
 
 ```{r}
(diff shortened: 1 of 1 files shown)


jcoliver:
Thanks, @cbeleites . I wonder if we might drop the trig functions entirely. One approach would be to use three different functions, `log`, `log10`, and `exp`, but use the same argument (`10`) in each example. The point of this super-duper introductory section is to provide some familiarity with how R deals with information, and whether a function assumes radians or degrees may be adding too much a load. Thoughts, @naupaka and @mawds ?

See also #430 .

cbeleites:
@jcoliver Thanks. Will add my thoughts at #430 to avoid parallel discussions.

jcoliver:
Closing this for now, but see discussion at #430 regarding the use of mathematical functions at this point of the lesson. We may end up dropping `sin`, `log`, etc. altogether to minimize cognitive load.

PR #471, opened 2019-02-06
Added missing dataset (feline-data_v2.csv) and adding zip folder to Setup page

sharonwsolis:
I added the missing dataset referred to in Episode 4, Data Structures. It originally says (around line 125 in .Rmd) "A user added details of another cat. This information is in the file data/feline-data_v2.csv." This was confusing because we had not yet created this file nor were we told to download the file. I added the file in the data folder and created a ZIP file to the Setup page to allow for easy download of all four files the lesson refers to.


naupaka:
Hi @sharonwsolis -- thanks for the PR. I think this is a rather substantial change to the flow of the lesson, since it has the learner creating a directory and etc. I agree that the mention of a dataset that doesn't exist and hasn't been created, but my inclination to solve that problem by removing that section/example altogether, instead of adding the complexity of another file... 

Thoughts @mawds @jcoliver ?

Also, generally we ask that PRs are only on the Rmd files, and don't include the rendered markdown, since these get regenerated by the maintainers and the makefile used to rebuild the lesson can sometimes behave unexpectedly if the timing of file creation/modification is non-standard.

naupaka:
I am going to close this PR, because I think the best solution is to remove the reference to this second csv altogether (and the inclusion of the rendered md in the PR makes it difficult to merge easily). We've added a new issue describing this problem (#479), which also links back here.

Issue #460, opened 2019-01-08
Section 7, Challenge 4 & 5 suggestions


jpiaskowski:
Challenges 4 & 5 are very difficult because the solutions expect four functions not previously introduced:

with()
subset()
mean()
cat()

The use of 'subset' is particularly confusing since the previous section is titled "Subsetting", yet does not introduce subset() at all. Is this intended for advanced users? I am skeptical that true R novices will be able to complete these exercises.

Here is the original solution:

`
for( iContinent in unique(gapminder$continent) ){
   tmp <- mean(subset(gapminder, continent==iContinent)$lifeExp)
   cat("Average Life Expectancy in", iContinent, "is", tmp, "\n")
   rm(tmp)
}
`

I think it can be completed using existing knowledge and only introducing mean() and cat() as hints. 

`
for( iContinent in unique(gapminder$continent) ){
  tmp <- gapminder[gapminder$continent == iContinent, ]
  cat(iContinent, mean(tmp$lifeExp, na.rm = TRUE), "\n")
  rm(tmp)
`

I would introduce na.rm = T since users are quite likely to eventually be tripped up by NA handling. 
I would also consider introducing cat in the very first section of the workshop as a "Hello World" incantation:

`
cat("Hello World")
`

jcoliver:
Great points, @jpiaskowski . Updating challenges 4 & 5 with solutions that do not involve `subset` and `with` sounds like a good idea.

jpiaskowski:
okay, I'll prep another pull request for this.

jcoliver:
Addressed in #472

PR #472, opened 2019-02-08
simplify 07-control flow exercises

jpiaskowski:
Hello, I made some changes to the exercises in  Episode 7 (control flow) per issue #460 that I had created. I introduced `cat()` earlier in the lesson so users would have seen it once already prior to the exercise. I left in the final use of `mean` and `subset` in final exercise thinking that if advanced users make it to that exercise, it will be a nice challenge.

### DIFF
--- a/_episodes_rmd/07-control-flow.Rmd
+++ b/_episodes_rmd/07-control-flow.Rmd
@@ -258,7 +258,7 @@ output_vector2
 >   do a thing
 > }
 > ```
-> R will interpret a condition being met as "TRUE". 
+> R will interpret a condition being met as "TRUE".
 >
 > As an example, here's a while loop
 > that generates random numbers from a uniform distribution (the `runif()` function)
@@ -268,7 +268,7 @@ output_vector2
 > z <- 1
 > while(z > 0.1){
 >   z <- runif(1)
->   print(z)
+>   cat(z, "\n")
 > }
 > ~~~
 > {: .r}
@@ -340,14 +340,14 @@ output_vector2
 > > We can do that as follows:
 > >
 > > 1. Loop over each of the unique values of 'continent'
-> > 2. For each value of continent, create a temporary variable storing the life exepectancy for that subset,
+> > 2. For each value of continent, create a temporary variable storing the that subset
 > > 3. Return the calculated life expectancy to the user by printing the output:
 > >
 > > ```{r 07-chall-03-sol-b, eval=FALSE}
 > > for( iContinent in unique(gapminder$continent) ){
-> >    tmp <- mean(subset(gapminder, continent==iContinent)$lifeExp)
-> >    cat("Average Life Expectancy in", iContinent, "is", tmp, "\n")
-> >    rm(tmp)
+> >   tmp <- gapminder[gapminder$continent == iContinent, ]   
+> >   cat(iContinent, mean(tmp$lifeExp, na.rm = TRUE), "\n")  
+> >   rm(tmp) }
 }
 > > ```
 > >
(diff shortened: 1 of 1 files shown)


_episodes_rmd/07-control-flow.Rmd
```diff
@@ -340,14 +340,14 @@ output_vector2
 > > We can do that as follows:
 > >
 > > 1. Loop over each of the unique values of 'continent'
-> > 2. For each value of continent, create a temporary variable storing the life exepectancy for that subset,
+> > 2. For each value of continent, create a temporary variable storing the that subset
```
naupaka: Looks like a minor typo here: "the that subset"

_episodes_rmd/07-control-flow.Rmd
```diff
@@ -340,14 +340,14 @@ output_vector2
...
 > > 3. Return the calculated life expectancy to the user by printing the output:
 > >
 > > ```{r 07-chall-03-sol-b, eval=FALSE}
 > > for( iContinent in unique(gapminder$continent) ){
-> >    tmp <- mean(subset(gapminder, continent==iContinent)$lifeExp)
-> >    cat("Average Life Expectancy in", iContinent, "is", tmp, "\n")
-> >    rm(tmp)
+> >   tmp <- gapminder[gapminder$continent == iContinent, ]   
+> >   cat(iContinent, mean(tmp$lifeExp, na.rm = TRUE), "\n")  
+> >   rm(tmp) }
```
naupaka: Closing `}` should go on its own line, per [tidyverse](https://style.tidyverse.org/syntax.html#indenting) and [Google R](https://google.github.io/styleguide/Rguide.xml#curlybraces) style guides (which we try to follow), so it looks like you have a duplicate closing `}` here.

_episodes_rmd/07-control-flow.Rmd
```diff
@@ -430,14 +430,14 @@ output_vector2
 > > ```
 > >
 > > We will now store these countries in a variable called candidateCountries, and then loop over each entry in the variable.
-> > Inside the loop, we evaluate the average life expectancy for each country, and if the average life expectancy is less than 50 we use base-plot to plot the evolution of average life expectancy:
+> > Inside the loop, we evaluate the average life expectancy for each country, and if the average life expectancy is less than 50 we use base-plot to plot the evolution of average life expectancy using `with` and `subset`:
```
naupaka: When describing functions we are trying to consistently include the parentheses -- so `with()` and `subset()` instead of the bare function names.

_episodes_rmd/07-control-flow.Rmd
```diff
@@ -430,14 +430,14 @@ output_vector2
...
 > >
 > > ```{r 07-chall-05-sol-c, eval=FALSE}
 > > thresholdValue <- 50
 > > candidateCountries <- grep("^B", unique(gapminder$country), value=TRUE)
 > >
 > > for( iCountry in candidateCountries){
-> >     tmp <- mean(subset(gapminder, country==iCountry)$lifeExp)
+> >     tmp <- mean(gapminder[gapminder$continent == iContinent, "lifeExp"])
```
naupaka: Isn't this going to be a problem because the loop is iterating over countries and not continents?

_episodes_rmd/07-control-flow.Rmd
```diff
@@ -390,7 +390,7 @@ output_vector2
 > >  upperThreshold <- 70
 > >  
 > > for( iCountry in unique(gapminder$country) ){
-> >     tmp <- mean(subset(gapminder, country==iCountry)$lifeExp)
+> >     tmp <- mean(gapminder[gapminder$continent == iContinent, "lifeExp"])
```
naupaka: Isn't this going to be an issue since the loop is iterating over countries not continents?

jpiaskowski: Good catch!

_episodes_rmd/07-control-flow.Rmd
jpiaskowski: Working on this today. FYI, There are a number of places in episode 7 where an R function is given and not followed by parentheses (e.g. lines 354 & 355- use of `if`, lines 418, 426,  use of `grep`. On line 339, `subset` is enclosed in backticks without parentheses, although in that case, it refers to the general concept of a data subset, not the actual R function. I fixed these when the term was clearly referring to a function.

jpiaskowski:
Okay, I think I properly created a pull request incorporating the suggested changes. Let me know if you do not see this. Thanks

naupaka:
Thanks!

PR #478, opened 2019-02-25
updated scale_x_log10 ggplot option to scale_x_continuous

ryandata:
Instead of scale_log_10 as an option to transform the plot to logarithmic scale in base 10, I suggest using scale_x_continuous(trans="log10").  Scale_x_continuous is equally intuitively understandable, and it allows the learner to easily use the same function for other transformations such as the natural log, exponential, and so on, without having to learn a new option.

### DIFF
--- a/_episodes_rmd/08-plot-ggplot2.Rmd
+++ b/_episodes_rmd/08-plot-ggplot2.Rmd
@@ -216,15 +216,10 @@ a large amount of data which is very clustered.
 
 ```{r axis-scale}
 ggplot(data = gapminder, aes(x = gdpPercap, y = lifeExp)) +
-  geom_point(alpha = 0.5) + scale_x_log10()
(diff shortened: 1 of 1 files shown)


naupaka:
Thanks for the PR! The maintainers will be meeting to deal with pull requests this coming Monday. We'll review this then and let you know if further changes are needed.

naupaka:
I'm going to close this PR, even though the maintainers discussed your suggestion and agree that this would be a positive change to the lesson. We have to close all open PRs in order to migrate the repo build system over to Travis CI from local builds. We have migrated the suggestion to an issue (#480). Please feel free to submit another PR to change this, but I'd also request that you don't turn off the evaluation of any of the chunks when you do so, and be mindful of the whitespace issues, since it makes diffing properly difficult.

PR #486, opened 2019-03-23
Rstudio changes

gauravsk:
I think some of the text in this lesson was written based on a previous version of the RStudio IDE. This Pull Request has two changes:
1. The new RStudio just has a "Knit" button, not "Knit HTML" as in the lesson.
2. There is no question mark next to the Knit button; instead, the Markdown reference is listed under the Help menu of the toolbar.  I also added text to point readers to the R Markdown Cheat Sheet, which is directly accessible from Rstudio

### DIFF
--- a/_episodes_rmd/15-knitr-markdown.Rmd
+++ b/_episodes_rmd/15-knitr-markdown.Rmd
@@ -178,10 +178,7 @@ with some number of `#` symbols:
 ```
 
 You _compile_ the R Markdown document to an html webpage by clicking
-the "Knit" in the upper-left. And note the little question mark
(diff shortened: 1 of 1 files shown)


naupaka:
Thanks for the PR, @gauravsk! Unfortunately the changes need to be against the master branch, since the gh-pages branch is now automatically generated via Travis-CI (recent change to our infrastructure, just happened in the last month). If you can resubmit these against master, I would be happy to consider merging. This should be as easy as selecting the proper branch when you open the PR.

gauravsk:
I followed this up with another PR (https://github.com/swcarpentry/r-novice-gapminder/pull/489), so closing this pull request

PR #489, opened 2019-03-26
Rstudio updates

gauravsk:
*Following up on a previous PR (https://github.com/swcarpentry/r-novice-gapminder/pull/486)*

I think some of the text in this lesson was written based on a previous version of the RStudio IDE. This Pull Request has two changes:

1. The new RStudio just has a "Knit" button, not "Knit HTML" as in the lesson.
2. There is no question mark next to the Knit button; instead, the Markdown reference is listed under the Help menu of the toolbar. I also added text to point readers to the R Markdown Cheat Sheet, which is directly accessible from Rstudio


naupaka:
Thanks @gauravsk !

Issue #221, opened 2016-10-10
Defensive Programming


rgaiacs:
Based on the discussion that I started at http://lists.software-carpentry.org/pipermail/discuss/2016-September/004778.html, can we include on this lesson the two episodes that John Blischak, @jdblischak, wrote Defensive programming with stopifnot
, https://jdblischak.github.io/r-intermediate-altmetrics/23-stopifnot.html, and Testing with testit, https://jdblischak.github.io/r-intermediate-altmetrics/24-testit.html?

mkcor:
Thanks, @rgaiacs .

:+1: to including lesson "Defensive programming with stopifnot" (editing reference to function `sum_calc_stat`). In a context of data analysis, I don't think I would agree with the statement that "it will involve more work" (since that's what you do anyway, you explore the data), but this is Software Carpentry, not Data Carpentry, so I won't argue.

I'm not as motivated to include lesson "Testing with testit" though.  I believe unit testing represents something more advanced (especially in the R world).  Plus it adds a dependency (be it `testit` or `testthat`).  Plus it touches on R packaging (as @lwjohnst86 rightly mentioned in the thread).

naupaka:
I am inclined to agree with @mkcor on this. I think the `stopifnot()` lesson would be a nice and appropriate addition, but I think that the lesson on unit tests might be a bit too much.

jdblischak:
I am defending next week, so I apologize in advance that I won't be able to have an extended discussion about this. Here are my thoughts:

*  I'm glad you liked my lessons! However, I am not sure how useful they will be as compared to just starting from scratch. The majority of the exposition and the examples refer to functions that were built earlier in the lesson.
*  I love using `stopifnot`. When I write code, I aim to take the quick sanity checks that I run in the R console and paste them into a `stopifnot` call. You can see my GitHub repos are [littered with stopifnot](https://github.com/search?o=desc&q=stopifnot+user%3Ajdblischak&ref=searchresults&s=&type=Code&utf8=%E2%9C%93).
[...]

PR #495, opened 2019-04-08
Replace stopifnot() with assert_that() for better error messages

katrinleinweber:
Hello :-)

What are your feeling/ideas/suggestions regarding replacing [`stopifnot()` in ep. 10](https://github.com/swcarpentry/r-novice-gapminder/blob/f3d5c959a39de2b4d9d9f6927523e8a969427004/_episodes_rmd/10-functions.Rmd) with [`assertthat::assert_that()`](https://github.com/hadley/assertthat#assertthat)?

This topic came up during a workshop preparation and is related to https://github.com/swcarpentry/r-novice-gapminder/issues/221. Over in [r-novice-inflammation](https://github.com/swcarpentry/r-novice-inflammation/search?q=stopifnot&unscoped_q=stopifnot), we don't teach it, and [previous plans to do it were decided against](https://github.com/swcarpentry/r-novice-inflammation/commit/a8a0c7326ea41aaae542806b15bb63c3bdecd3bb).

This is a quick draft, feedback welcome!

### DIFF
--- a/_episodes_rmd/01-rstudio-intro.Rmd
+++ b/_episodes_rmd/01-rstudio-intro.Rmd
@@ -590,6 +590,7 @@ network). R and RStudio have functionality for managing packages:
 > > install.packages("ggplot2")
 > > install.packages("plyr")
 > > install.packages("gapminder")
+> > install.packages("assertthat")
 > >```
 > {: .solution}
 {: .challenge}
--- a/_episodes_rmd/10-functions.Rmd
+++ b/_episodes_rmd/10-functions.Rmd
@@ -225,6 +225,7 @@ We want to assert the following: `temp` is a numeric vector. We may do that like
 so:
 
 ```{r}
+library(assertthat)
 fahr_to_kelvin <- function(temp) {
   assert_that(is.numeric(temp))
   kelvin <- ((temp - 32) * (5 / 9)) + 273.15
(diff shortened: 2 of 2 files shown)


naupaka:
We had a discussion about this a few years ago when `stopifnot()` was added (https://github.com/swcarpentry/r-novice-gapminder/issues/221). The sense then was that it was probably better to stick to a simple base R approach to emphasize defensive programming, instead of adding in another dependency since introducing the concept of unit tests (i.e. `testthat`) in reasonable detail would add a lot of content to already full materials... But perhaps `assertthat()` is reasonable update? @jcoliver @mawds?

jcoliver:
The value of `assertthat` is in the error messages:

`stopifnot`:
> Error: is.numeric(temp) is not TRUE

`assertthat`:
> Error: temp is not a numeric or integer vector

The syntax between the two is identical, so my question is: does the slightly more informative error message justify the cost of the added dependency?

katrinleinweber:
An additional benefit is the reminder to load the `library()` command. That plus the better readability of errors weighs more, than any risk this "dependency" might introduce. What could the risk be _at worst_? We're not running a commercial web app here, where each hour downtime causes lost profits ;-)

If this changes proves useful, we could later turn it into an opportunity to introduce the `::` operator, or show how `install.packages()` can handle a vector.

naupaka:
@katrinleinweber we've (cc @jcoliver @mawds) discussed this and would like to add it as a tip instead of changing the code in the lesson. It is a better function than `stopifnot()`, but the improvement is not so dramatic to overcome the downside of having another dependency (with the many lessons maintained by the Carpentries and all the automated build systems and different versions of R on Travis, locally and etc flying around it's best to avoid additional complexity when possible. 

Would you be willing to update the PR to make it a callout tip instead?

jcoliver:
We're going to close this for now, but if there is an interest in adding `assertthat()` as a callout tip, please feel free to re-open.

PR #504, opened 2019-05-01
Canonicalize CRAN links

katrinleinweber:
CRAN [asks people](https://cran.r-project.org/doc/manuals/R-exts.html#Specifying-URLs) to use this URL variant when linking to packages ;-) This PR results from a semi-automatic search-and-replace script and implements their suggestion.

### DIFF
--- a/_episodes_rmd/05-data-structures-part2.Rmd
+++ b/_episodes_rmd/05-data-structures-part2.Rmd
@@ -259,7 +259,7 @@ gapminder <- read.csv("data/gapminder_data.csv")
 > ```
 >
 > * You can read directly from excel spreadsheets without
-> converting them to plain text first by using the [readxl](https://cran.r-project.org/web/packages/readxl/index.html) package.
+> converting them to plain text first by using the [readxl](https://cran.r-project.org/package=readxl) package.
 {: .callout}
 
 Let's investigate gapminder a bit; the first thing we should always do is check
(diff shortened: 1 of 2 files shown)


jcoliver:
Thanks @katrinleinweber ! Looks good. Will merge pending successful Travis CI build.

jcoliver:
@katrinleinweber thanks again for this. Would you update the PR so only the .Rmd files (and not the .md) files are changed?

katrinleinweber:
Do the `.md` changes break Travis' build of `gh-pages`? Since they introduce the exact same changes as in the `.Rmd` files I assumed they would merge & build cleanly.

jcoliver:
@katrinleinweber I _think_ so, but I'll defer to @naupaka . What sort of issues do edited .md files cause during build?

naupaka:
Depends on the modification times because we rely on a makefile. Easiest to just delete the `.md` files so there's no possibility of it thinking it's already built, esp. given merging of other PRs, etc.

katrinleinweber:
I assumed that the modification times of all files become the timestamp of Travis cloning the git repo into its VM. Thus, the makefile would see the same timestamp for both `.md` & `.Rmd` files anyway.

jcoliver:
Thanks, @katrinleinweber!

katrinleinweber:
You're welcome :-)

naupaka:
Looks great, thanks!

Issue #502, opened 2019-04-29
Remove color by continent at beginning of Transformations & Statistics (08)


jcoliver:
At the beginning of "Transformations & Statistics" section of lesson 08 (ggplot), the example has `color = continent`. Subsequent code blocks do not have this aesthetic set, and a challenge later on assigns this to users. The `color = continent` should be removed.

naupaka:
Closed via #505

PR #505, opened 2019-05-02
Remove color from Transformations & statistics (08)

jcoliver:
Per #502 , removing `color=continent` at beginning of Transformations & Statistics section in order to maintain continuity with subsequent portions of the lesson.


naupaka:
Looks good, thanks!

PR #509, opened 2019-05-28
Added links to packages used in the lesson

wna-se:
Listed the packages that will be downloaded and installed during the lesson to the setup inatructions. This is helpful in situations where IT staff need to be informed about the packages installed on the computers.


jcoliver:
Thanks, @wna-se

Issue #480, opened 2019-03-04
Change scale_x_log10 ggplot option to scale_x_continuous in episode 8


jcoliver:
In ggplot episode, per #478, use `scale_x_continuous` with the argument `trans="log10"` instead of the shortcut `scale_x_log10` function.

PR #519, opened 2019-06-06
Change scale_x_log10 ggplot option to scale_x_continuous in episode 8 #480

waiteb5:
Checkout and first addition. Attempting to address issue https://github.com/swcarpentry/r-novice-gapminder/issues/480


jcoliver:
Thanks, @waiteb5 !

Issue #479, opened 2019-03-04
Remove feline-data_v2.csv


jcoliver:
Per #471, consider removing all references to the cats version 2 data.

anuj2054:
Maybe just adding this code in the section where it asks us to download the file might help. 
Or I can go ahead and delete the whole section.

```
cats <- data.frame(coat = c("calico", "black", "tabby","white"), 
                    weight = c(2.1, 5.0,3.2,2.1 or 2.2), 
                    likes_string = c(1, 0, 1,1))
write.csv(x = cats, file = "data/feline-data_v2.csv", row.names = FALSE)
```

dkcoxie:
Taught this lesson for the first time and found the feline_v2.csv reference very confusing. I think it should removed. If a 'badly' behaved data frame is included, I think it would fit better as an example in an example in 02-project-intro, as something that can be downloaded, changed in a text editor, and then loaded into R to demonstrate the how best practices of project organization/data analysis can help them avoid errors/issues.

jcoliver:
We are making significant revisions to episodes 4 and 5 due to changes introduced by R 4.0. Please feel free to re-open as warranted.

ericward:
This still seems not to have been addressed?

PR #533, opened 2019-07-03
Fix feline v2.csv in data structures

madcowen:
This commit has the user create a csv file (feline-v2.csv), which is used in the data structures episode but never made.


naupaka:
Hi @madcowen thanks for the PR. I believe our approach to solving this was discussed in #479, which is to remove the references to the cats v2 altogether to simplify the lesson

jcoliver:
This episode is undergoing revisions per the update to R 4.0 (#633), so we are closing this issue for now. Feel free to re-open if warranted.

PR #536, opened 2019-07-05
Fixed broken link caused by misplaced backticks

shihchingfu:
In Rmarkdown, backticks specifying inline-code formatting take precedence over square brackets used for inserting a hypertext link. Thus the backticks should be inside the square brackets rather than around the entire statement.

Otherwise the knitted document shows the literal string instead of being a clickable link.


jcoliver:
Thanks, @shihchingfu !

PR #538, opened 2019-07-14
Add for loop example for comparison with vectorized code

### LINKED ISSUE #523: episode 9 (vectorizing): compare with for loop output?
Hi Carpentries folks,

The listed [key point in episode 9 about vectorizing](https://github.com/swcarpentry/r-novice-gapminder/blob/2169686b07a57e46821a0a1351e9fba2e2a8691f/_episodes_rmd/09-vectorization.Rmd#L10) is to
> "Use vectorized operations instead of loops."

which is great! However, in the episode, there are no side-by-side examples of accomplishing the same task using both a vectorized call and a for loop. The first piece of explanatory text in the lesson plan [explicitly compares vectorizing with for looping:](https://github.com/swcarpentry/r-novice-gapminder/blob/2169686b07a57e46821a0a1351e9fba2e2a8691f/_episodes_rmd/09-vectorization.Rmd#L23)

> ... without needing to loop through
> and act on each element one at a time. This makes writing code more
> concise, easy to read, and less error prone.

This could couple really well with a worked example showing a multi-line for loop that loops over the indices of a vector, and contrasting it with a single-line vectorized call that has equivalent output. That might serve to make very concrete that vectorized functions are often less verbose than equivalent for loops. Learners coding along would likely find that typing the vectorized call is much faster and less typo-prone than typing out the for loop, thus hitting the point home. 

Since this episode comes after the control flow episode, presumably learners will have seen a for loop before, so including a for loop in the worked code hopefully wouldn't take too much time for explanation.

I know adding worked code examples to episodes is tough since they're already so info-packed, but this might be able to be substituted for the chunk explaining `%*%` that starts [here.](https://github.com/swcarpentry/r-novice-gapminder/blob/2169686b07a57e46821a0a1351e9fba2e2a8691f/_episodes_rmd/09-vectorization.Rmd#L142) (People coming from other not-super-vectorized languages like Matlab probably benefit from being shown that the matrix multiplication operator is not the default in R, but this may not be a super common use case that needs a worked code example in this intro-level lesson.)

happy to discuss and maybe to try suggesting edits in a PR, but figured opening an issue was a good place to start.

thanks folks!

(PS: I am doing this as part of instructor training checkout)

fjuniorr:
Resolves #523 .

Ps 1. This is my first PR. I'm not sure how much from the discussion on the issue I should duplicate in the PR.

Ps 2. Should I commit the changes to the `.md` file as well?

### DIFF
--- a/_episodes_rmd/09-vectorization.Rmd
+++ b/_episodes_rmd/09-vectorization.Rmd
@@ -49,6 +49,15 @@ y:  6  7  8  9
     7  9 11 13
 ```
 
+Here is how we would add two vectors together using a for loop:
+
+```{r}
+output_vector <- c()
+for(i in 1:4) {
+  output_vector[i] <- x[i] + y[i]
+}
+output_vector
+```
 
 > ## Challenge 1
 >
--- a/_episodes_rmd/09-vectorization.Rmd
+++ b/_episodes_rmd/09-vectorization.Rmd
@@ -57,6 +57,9 @@ for(i in 1:4) {
   output_vector[i] <- x[i] + y[i]
 }
 output_vector
+
+sum_xy <- x + y
+sum_xy
 ```
 
 > ## Challenge 1


monicathieu:
What you have looks like a great start! I think per what @jcoliver mentioned in issue #523, you might also amend Challenge 1 below what you've added to ask both for a for-loop answer and a vectorized answer (if there's time in the lesson). That way learners can reeeeally see (both with the example and with self-generated challenge answering code) that in most cases, vectorized solutions do the same thing as for loop solutions but with less code.

jcoliver:
Thanks, @fjuniorr . I think adding the call to `sum_xy` assignment and printing helps reinforce the concept (and instructors can skip the `sum_xy` if they choose). Want to try adding that in?

In regards to pull requests, especially if they reference open issues, no need to duplicate the entirety of the rationale behind the PR. A brief description is sufficient. Often times a descriptive commit comment is enough. And don't worry about the `.md` files - those are automatically generated from the `.Rmd` files when we update the lesson.

I agree with @monicathieu that adding a vector answer to Challenge 1 is a good idea. However, it need not be addressed with this PR. We can leave issue #523 open and update it with the work that still needs to be done.

jcoliver:
Looks great @fjuniorr . After the automated checks complete, we'll merge the PR. Thanks again.

PR #546, opened 2019-08-04

seacatkim:
Expanding on matrices description distinguishing key difference from data frames:

"Last but not least is the matrix. Similar to data frames, matrices are two dimensional with rows and columns. A data frame can combine columns of different data types such as integers, factors, logical etc. while matrices are homogenous only containing a single data type for the whole matrix.

We can declare a matrix full of zeros:"


jcoliver:
Thanks, @seacatkim !

PR #547, opened 2019-08-12
Update lessons 4 13 14

annesbe:
Please delete this line and the text below before submitting your contribution.

---

---


jcoliver:
Thank you for the PR, @annesbe . While including `row.names = FALSE` and `na.rm = TRUE` can be useful parameters, I wonder if we want to add them at this part of the lesson. Some of us use these in our routine workflow, but adding them at this point would require additional explanation and, especially in the tidyr lesson, a fair amount of additional typing on our learners' part.

naupaka:
@jcoliver I am in agreement on this. I don't think there's a major benefit to including these additional parameters in this location in the lesson, and there may be some downsides. While they are often best practice in day-to-day work, we need to be careful to limit cognitive load in these lessons and the data we are using here are clean and so it is not necessary.

I am also not sure why it's necessary to remove the sorting by continent?

annesbe:
Thanks for the feedback, guys! I will update the PR asap.

PR #549, opened 2019-08-12
Simplifying the section on Multi-panel figures

rwmurdoch:
for creating a multi-panel figure, the data is subsetted to only those beginning with A or Z.  The method applied is creative but very challenging!  It doesn't add much to the exercise, which is obviously focused on the capabilities of ggplot2 rather than interesting shortcuts for data subsetting.  I'm proposing a simple subsetting based on matching one of the variables (continent), a strategy which is applied earlier in the workshop.

### DIFF
--- a/_episodes_rmd/08-plot-ggplot2.Rmd
+++ b/_episodes_rmd/08-plot-ggplot2.Rmd
@@ -315,7 +315,9 @@ by adding a layer of **facet** panels.
 ```{r facet}
 americas <- gapminder[gapminder$continent == "Americas",]
 ggplot(data = americas, mapping = aes(x = year, y = lifeExp)) +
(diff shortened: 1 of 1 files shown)


jcoliver:
Cool @rwmurdoch ! Could you update the PR with the ggplot functions separated to one line each. i.e. change 
```
  geom_line() + facet_wrap( ~ country) + theme(axis.text.x = element_text(angle = 45))
```
to
```
  geom_line() + 
  facet_wrap( ~ country) + 
  theme(axis.text.x = element_text(angle = 45))
```

rwmurdoch:
Good point, I cleaned up the script

jcoliver:
Excellent. Thanks, @rwmurdoch .

Issue #542, opened 2019-07-30
Suggestion to add a gentler introduction for ggplot2


wallacelab:
The current [ggplot2 lesson](http://swcarpentry.github.io/r-novice-gapminder/08-plot-ggplot2/index.html) feels like it starts very abruptly, going from nothing to using ggplot(), aes(), and geom_point() all in a single line of code. That seems like a bit much for a first experience. (I learned ggplot2 ~6 months ago through this very lesson, and I felt very lost until I could go back through a second or third time.)

It seems like a little expansion of the introduction would help with this. The lesson currently has these lines:

> ggplot2 is built on the grammar of graphics, the idea that any plot can be expressed from the same set of components: a data set, a coordinate system, and a set of geoms–the visual representation of data points.
> 
> The key to understanding ggplot2 is thinking about a figure in layers. This idea may be familiar to you if you have used image editing programs like Photoshop, Illustrator, or Inkscape.

I suggest expanding/altering it into something more like

> ggplot2 is built on the grammar of graphics, the idea that any plot can be expressed from the same set of components. These components are a data set, mapping aesthetics, and layers of graphics.
> - **Data sets** are the data that you, the user, provide.
> - **Mapping aesthetics** are what connect the data to the graphics. They tell ggplot2 how to use different parts of your data to affect how the graph looks (such as by changing what is plotted on the X or Y axis, or if parts of the plot should have different sizes or colors based on your data).
> - **Layers** are the actual graphical output from ggplot2. Layers determine what kind(s) of plot are shown (scatterplot, histogram, etc.), the coordinate system used (rectangular, polar, etc.), and other important aspects of the plot. The idea of layers of graphics may be familiar to you if you have used image editing programs like Photoshop, Illustrator, or Inkscape.

When teaching I like diagramming these out, but I don't know if that belongs in the official lesson itself. Same for pointing out these three components in the first code example.

The reasons I suggest replacing "coordinate system" with "mapping aesthetics" are because
- The word "coordinate" is never used again in the lesson, whereas "mapping" shows up 25 times and "aesthetic" 16 times 
- None of the examples actually change the coordinate system. (The closest is adjusting the scale with `+ scale_x_log10()` )
- The base ggplot() command seems to function more like a "layer 0" than a coordinate system. (This may be due to an incorrect mental map on my part, but it behaves like the base layer on which everything else is placed.)
- This would make clearer what aes() is actually doing (mentioned briefly in issue #473)

[...]

Issue #473, opened 2019-02-12
Specify `mapping` in the ggplot2 episode?


ha0ye:
I like the approach of the data carpentry R ecology lesson where the `mapping` named argument is explicitly specified for ggplot2. This assists with distinguishing the different layers and maybe also with explaining what's occurring under the hood (rather than a "magical" `aes()` function).

mawds:
It's a good point, and would, I think make it clearer what the "point" of `aes()` is, especially at the start of the episode.  I know I emphasise that `aes()` is a "mapping" when I'm teaching it, and I think this would make it clearer.

naupaka:
I agree with this too. When people are learnings, it's always helpful to be explicit at first with the naming of arguments.

jcoliver:
Resolved with #492

PR #552, opened 2019-08-14
Make gentler intro for ggplot2 lesson

wallacelab:
This is in response to Issue #542. The current ggplot2 lesson starts very abruptly and can be confusing to novices. In my teaching, I've found that using a little time to explain the logic of ggplot a bit more helps a lot. I've also found that building up the first graphic piece by piece works better than starting with the finished graphic and then backtracking to explain how it fits together. 

All changes in this pull request are in the first section of the lesson (up to but not including the first challenge). There are two main (and separate) changes that have been made:
1. The introduction to ggplot grammar has been expanded to replace "coordinate system" with "mapping aesthetics" and explain in (slightly) more detail what each component is.
2. The example scatterplot is built up one element at a time (first data, then aesthetics, then geom) rather than all at once.

In my experience these changes make it much easier for learners to understand how ggplot works. The major disadvantage is that it probably adds ~5 minutes of instruction time.

### DIFF
--- a/_episodes_rmd/08-plot-ggplot2.Rmd
+++ b/_episodes_rmd/08-plot-ggplot2.Rmd
@@ -40,52 +40,66 @@ it is the most effective for creating publication quality
 graphics.
 
 ggplot2 is built on the grammar of graphics, the idea that any plot can be
-expressed from the same set of components: a **data** set, a
-**coordinate system**, and a set of **geoms**--the visual representation of data
-points.
+built from the same set of components: a **data set**,
+**mapping aesthetics**, and graphical **layers**:
 
-The key to understanding ggplot2 is thinking about a figure in layers.
-This idea may be familiar to you if you have used image editing programs like Photoshop, Illustrator, or
-Inkscape.
+* **Data sets** are the data that you, the user, provide.
 
-Let's start off with an example:
+* **Mapping aesthetics** are what connect the data to the graphics.
+They tell ggplot2 how to use your data to affect how the graph looks,
+such as changing what is plotted on the X or Y axis, or the size or
+color of different data points.
 
-```{r lifeExp-vs-gdpPercap-scatter, message=FALSE}
-library("ggplot2")
-ggplot(data = gapminder, mapping = aes(x = gdpPercap, y = lifeExp)) +
-  geom_point()
-```
+* **Layers** are the actual graphical output from ggplot2. Layers
+determine what kinds of plot are shown (scatterplot, histogram, etc.),
+the coordinate system used (rectangular, polar, others), and other
+important aspects of the plot. The idea of layers of graphics may
+be familiar to you if you have used image editing programs
+like Photoshop, Illustrator, or Inkscape.
 
-So the first thing we do is call the `ggplot` function. This function lets R
-know that we're creating a new plot, and any of the arguments we give the
-`ggplot` function are the *global* options for the plot: they apply to all
+Let's start off building an example using the gapminder data from earlier.
+The most basic function is `ggplot`, which lets R know that we're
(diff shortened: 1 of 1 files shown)


_episodes_rmd/08-plot-ggplot2.Rmd
```diff
@@ -40,53 +40,67 @@ it is the most effective for creating publication quality
...
-expressed from the same set of components: a **data** set, a
-**coordinate system**, and a set of **geoms**--the visual representation of data
-points.
-
-The key to understanding ggplot2 is thinking about a figure in layers.
-This idea may be familiar to you if you have used image editing programs like Photoshop, Illustrator, or
-Inkscape.
-
-Let's start off with an example:
+built from the same set of components: a **data set**,
```
wallacelab: Change #1: Expand the explanation of how ggplot works to make it easier for novices to understand. Also replaced "coordinate system" with "mapping aesthetics" for the following reasons (from issue #542)
- The word "coordinate" is never used again in the lesson, whereas "mapping" shows up 25 times and "aesthetic" 16 times
- None of the examples actually change the coordinate system. (The closest is adjusting the scale with + scale_x_log10() )
- The base ggplot() command seems to function more like a "layer 0" than a coordinate system. (This may be due to an incorrect mental map on my part, but it behaves like the base layer on which everything else is placed.)
- This would make clearer what aes() is actually doing (mentioned briefly in issue #473)

_episodes_rmd/08-plot-ggplot2.Rmd
```diff
@@ -40,53 +40,67 @@ it is the most effective for creating publication quality
...
+such as changing what is plotted on the X or Y axis, or the size or
+color of different data points.
+
+* **Layers** are the actual graphical output from ggplot2. Layers
+determine what kinds of plot are shown (scatterplot, histogram, etc.),
+the coordinate system used (rectangular, polar, others), and other
+important aspects of the plot. The idea of layers of graphics may
+be familiar to you if you have used image editing programs
+like Photoshop, Illustrator, or Inkscape.
+
```
wallacelab: Second major change (below): Broke the first example apart so that it's built up in stages (data, aesthetics, geoms) rather than putting in a large block of code all at once. In my experience this helps learners understand better what each component is doing and how they fit together.

_episodes_rmd/08-plot-ggplot2.Rmd
```diff
@@ -334,11 +348,11 @@ elements. The x-axis is too cluttered, and the y axis should read
 "Life expectancy", rather than the column name in the data frame.
```
wallacelab: Not sure how the below edits happened since they didn't show up in my initial comparison. Regardless, they're just spaces at the end of the line and can be ignored.

jcoliver:
Thanks for the PR, @wallacelab . We're reviewing the changes.

naupaka:
@wallacelab this seems like a nice PR but it's currently got a merge conflict against main. If you can make the adjustments necessary for a clean merge, I think we'd be happy to merge this in. We need to merge or close all open PRs before the end of the month (April 2023) since the primary branch structure will be changing and all existing PRs will break.

matthieu-bruneaux:
Thank you @wallacelab and @naupaka!

I fixed the merge conflicts from this PR using the command line (locally on my computer), and I then created a new PR (#826) that should allow for a clean merge of the content of this PR (#552).

This is my first time resolving merge conflicts via the command line and pushing back to the GitHub repo for a clean merge, please let me know if there was a simpler/preferred way to do this :)

naupaka:
Thanks @matthieu-bruneaux! I'm closing this PR and we can merge the new one.

PR #554, opened 2019-08-21
corrected inconsistent spacing in weight vector

geekypete:
Corrected inconsistent spacing weight vector of cats dataframe so the weight vector is spaced consistently with the coat and likes_string vectors.


jcoliver:
Thanks for the PR, @geekypete .

PR #557, opened 2019-08-27
Update 02-project-intro.md

Cecilia-Wang:
Added keyboard method to save a file with macOS @ line 151.


Cecilia-Wang:
Have changed and committed using the Rmd file instead.

PR #563, opened 2019-09-28
spacing, ifelse, and vectorized changes 

atheobold:
Added consistent spacing, explanation of ifelse error based on length of input vector, and advice on what to use if iteration order doesn't matter.


jcoliver:
Thank you for the PR, @atheobold ! These are great clarifications. Could you update the PR and put the for loop syntax lead-in back? It's line 173 of the current version:

The basic structure of a `for()` loop is:

Which should appear right before
```{r, eval = FALSE}
for(iterator in set of values){
  do a thing
}
```

jcoliver:
@fmichonneau this PR seems like the Travis build should be finished (it exited with status 0 according to the log). Any ideas why it's listed as not being completed?

fmichonneau:
I'm not sure, it looks like a fluke with the API. I think it can be merged.

jcoliver:
Thanks, @atheobold !

PR #564, opened 2019-09-30
Correct sub- & superscript formatting

katrinleinweber:
This is more of an issue/bug-report, but because of a related fix.

The sub- & superscript examples seem to [not be rendered online](https://swcarpentry.github.io/r-novice-gapminder/15-knitr-markdown/index.html#a-bit-more-markdown), but are rendered when using RStudio's knitr button.

Thus our code is fine here (with the fix), but the lesson template somehow doesn't render it.

### DIFF
--- a/_episodes_rmd/15-knitr-markdown.Rmd
+++ b/_episodes_rmd/15-knitr-markdown.Rmd
@@ -222,7 +222,7 @@ You can make a hyperlink like this:
 
 You can include an image file like this: `![caption](http://url/for/file)`
 
-You can do subscripts (e.g., F~2~) with `F~2` and superscripts (e.g.,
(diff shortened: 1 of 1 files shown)


jcoliver:
The same is true for LaTeX formatting immediately succeeding sub/superscripts (they aren't rendered correctly online either). Definitely a bug.

fmichonneau:
the issue is that MathJax is not part of our template, see https://github.com/carpentries/styles/issues/154

katrinleinweber:
Ah, OK! Thanks for explaining :-) In that case, this PR can be about the typo only.

jcoliver:
Thanks for the PR, @katrinleinweber , and thanks for the explanation, @fmichonneau . I'll open a separate issue for the math formatting.

katrinleinweber:
You're welcome, and thanks for merging :-)

PR #569, opened 2019-10-07
Separate R and RStudio download links with unordered bullets

matt-dray:
The links for downloading R and RStudio appear on the same line with no punctuation between them In the [Before Starting the Workshop](http://swcarpentry.github.io/r-novice-gapminder/01-rstudio-intro/#before-starting-the-workshop) section of the first lesson.

I think this could be confusing for users since it looks like one link where there's actually two. Here's a screenshot of the problem with the two links underlined separately in red:

Screenshot 2019-10-07 at 13 09 31

I think the intention was probably to have these on separate lines or as bullets, since they appear on separate lines in the R Markdown file.

This pull requests adds unordered bullets to separate these instructions onto separate lines. The output looks like this:

Screenshot 2019-10-07 at 13 22 21

I've checked the issues and PRs for this problem, but didn't find anything.


jcoliver:
Thanks, @matt-dray !

PR #588, opened 2019-11-14
challange 5 - boxplot

simonezuffa:
Change challenge 5 from density plot to boxplot


jcoliver:
Thanks @simonezuffa !

PR #593, opened 2019-11-25
Update ggplot2 docu & cheatsheet links

katrinleinweber:
See diff ;-) These are the current links that RStudio also uses.


jcoliver:
Thanks (again!) @katrinleinweber !

katrinleinweber:
You're welcome & thanks for merging :-)

PR #597, opened 2019-11-28
Update 03-seeking-help.Rmd

marklcrowe:
Challenge 3 describes a "csv" file delimited by the tab character. By definition, a csv file is comma-delimited. I'd like to propose replacing references to csv with table/tabular to remove this inconsistency

### DIFF
--- a/_episodes_rmd/03-seeking-help.Rmd
+++ b/_episodes_rmd/03-seeking-help.Rmd
@@ -168,12 +168,12 @@ your issue.
 
 > ## Challenge 3
 > Use help to find a function (and its associated parameters) that you could
-> use to load data from a csv file in which columns are delimited with "\t"
+> use to load data from a tabular file in which columns are delimited with "\t"
 > (tab) and the decimal point is a "." (period). This check for decimal
 > separator is important, especially if you are working with international
 > colleagues, because different countries have different conventions for the
 > decimal point (i.e. comma vs period).
-> hint: use `??csv` to lookup csv related functions.
+> hint: use `??"read table"` to look up functions related to reading in tabular data.
 > > ## Solution to Challenge 3
 > >
 > > The standard R function for reading tab-delimited files with a period


_episodes_rmd/03-seeking-help.Rmd
```diff
@@ -168,12 +168,12 @@ your issue.
...
 > ## Challenge 3
 > Use help to find a function (and its associated parameters) that you could
-> use to load data from a csv file in which columns are delimited with "\t"
+> use to load data from a tabular file in which columns are delimited with "\t"
 > (tab) and the decimal point is a "." (period). This check for decimal
 > separator is important, especially if you are working with international
 > colleagues, because different countries have different conventions for the
 > decimal point (i.e. comma vs period).
-> hint: use `??csv` to lookup csv related functions.
+> hint: use `??"read table"` to look up functions related to reading in tabular data.
```
jcoliver: `??read table` throws an error due to the space between "read" and "table" (`Error: unexpected symbol in "??read table"`). This statement should be revised.

jcoliver:
Great point, @marklcrowe . Want to take a crack at the suggested revision?

_episodes_rmd/03-seeking-help.Rmd
marklcrowe: Hi @jcoliver, thanks for the feedback. You're right, `??read table` gives an error, but `??"read table"` should work. Do I need to escape the double-quotes, or did you just overlook them when checking?

jcoliver: Yar! I totally overlooked the double-quotes. _But_, when I run `??"read table", no results are returned. Should there be?

marklcrowe: I've just tested it on a completely fresh install of R (current release - 3.6.1) and it comes up with base::base-defunct and utils::read.table - that's the same whether I'm using command line, the R app, or RStudio (on MacOS Mojave). Are you testing through RStudio? I know that for at least one of the recent RStudio releases, the "??" function didn't work at all.

jcoliver: I can confirm this is an RStudio problem (https://stackoverflow.com/questions/57623417/double-question-mark-help-search-not-working-in-rstudio-and-rstudio-only).

Thanks for testing this out, @marklcrowe!

Issue #572, opened 2019-10-17
File missing in lesson 4 of R for reproducible scientific analyses


SaraTomiolo:
Hi, I am going through the lesson number 4 of [R for reproducible scientific analyses](https://swcarpentry.github.io/r-novice-gapminder/04-data-structures-part1/index.html). In the section of data types there is a prompt to load the file "feline-data_v2.csv" from the data folder, but that file does not exist.
It seems that the file may have been modified from the original "feline-data.csv" but there is no indication in the prior parts of the lesson of that happening.

SaraTomiolo:
Hi, I have tried to look at the R Markdown file of the lesson. I think the easiest way to fix this would be to provide the user with a link (as is done for gapminder) to download this "feline-data_v2.csv" data set. Unfortunately I am not advanced enough to understand how to fix this. Sorry!

mstrimas:
Based on this older [issue](https://github.com/swcarpentry/r-novice-gapminder/issues/479) and [PR](https://github.com/swcarpentry/r-novice-gapminder/pull/533), it seems the intention is to actually remove this example altogether. Could be worth closing this issue so other trainees going through checkout process don't accidentally try to address this issue incorrectly.

mstrimas:
FWIW I think the entire feline v2 example is confusing and out of place. I agree that removing the whole thing would be an improvement. The section on Type Conversion seems to cover similar concepts in a less confusing fashion.

jcoliver:
Closing, as issue #479 supersedes.

PR #602, opened 2020-01-11

mstrimas:
The file `data/feline-data_v2.csv` is referenced in text and used in the examples, but there is no link given to download the file. This commit adds a link to the file and download instructions. This addresses issue swcarpentry/r-novice-gapminder#572

I also find it odd that `data/feline-data.csv` is created via code at the top of this lessons. Seems like a waste of time typing that in. Why not have folks download this file as well? If there's interest I could add that to this commit or a new one.

### DIFF
--- a/_episodes_rmd/04-data-structures-part1.Rmd
+++ b/_episodes_rmd/04-data-structures-part1.Rmd
@@ -123,8 +123,7 @@ complicated our analyses become, all data in R is interpreted as one of these
 basic data types. This strictness has some really important consequences.
 
 A user has added details of another cat. This information is in the file
-`data/feline-data_v2.csv`.
-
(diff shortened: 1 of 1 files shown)


mstrimas:
Oh, I see now from [this issue](https://github.com/swcarpentry/r-novice-gapminder/issues/479) and [this PR](https://github.com/swcarpentry/r-novice-gapminder/pull/533) that the intention is to actually remove this example using feline data v2 altogether, so this PR is obsolete. It's a bit confusing because [this more recent issue](https://github.com/swcarpentry/r-novice-gapminder/issues/572) suggests adding a download link, which is what I did.

jcoliver:
Thanks, @mstrimas . I agree that feline data v2 may be deprecated in the future, but until that time, the link may be useful for some instructors.

PR #608, opened 2020-01-18
Update 01-rstudio-intro.Rmd

MattGurney:
Show have multiple packages can be installed in a single install.packages() command


jcoliver:
Cool! Thanks, @MattGurney .

PR #628, opened 2020-04-24
Rendering problem

AnanthanAmbikairajah:
The file was not rendering because the same chunk header 'ch5-sol' was used twice. I updated this so that there were unique chunk headers for the file to render. I also added another alternative solution to challenge number 5


jcoliver:
Thanks, @AnanthanAmbikairajah !

jcoliver:
This rendering problem appears solved due to other changes.

Issue #623, opened 2020-03-13
Episode 1 Intro to R & RStudio, managing environment 


apculley:
I’d suggest following the sentence: ‘If we type ls by itself, R will print out the source code for that function!’ with a very brief layperson explanation of what source code is, such as: ‘Source code what is going on behind the scenes whenever a function is run.’ (And then maybe add that we’ll be going over functions in detail later).

jcoliver:
Thanks, @apculley !

blongworth:
I've seen a lot of R learners forget the parens, especially if they've just taken a shell lesson. It's also a great time to introduce thinking of everything in R as an object, and to touch on default methods, in this case print().

PR #629, opened 2020-04-29
Explain object printing behavior

blongworth:
Added an explanation of why `ls` without parens prints the contents of the ls object.

This should fix issue #623 

`make serve` didn't want to render until I renamed a duplicate code block in one of the challenge solutions.


jcoliver:
Thanks, @blongworth !

Issue #581, opened 2019-10-29
Unclear diagrams


mattgillucl:
On the page [_R for Reproducible Scientific Analysis - Dataframe Manipulation with dplyr_](https://swcarpentry.github.io/r-novice-gapminder/13-dplyr/index.html), there is a diagram with the heading `gapminder %>% group_by(a)`.

In the diagram, it is unclear whether or not the top two arrows cross. Please change this diagram to make it clear.

mattgillucl:
PS The next diagram below this one has the same problem.

jcoliver:
I agree. The arrows should be re-drawn and potentially re-colored to improve utility of these graphics.

jrmuirhead:
I've submitted a pull request changing the R code to generate the figures. One question I had @jcoliver , are the png versions in the fig directory generated as part of the TravisCI checks, or will I have to update the png files as well?

jcoliver:
Code updated (#631 ) and png regenerated (#644 ).

PR #631, opened 2020-05-02
Issue #581 fix. Figures redrawn

jrmuirhead:
- Figures were redrawn to avoid overlapping arrows to eliminate potential confusion.
- Colors were also changed in order to be colorblind safe.


jcoliver:
Thanks, @jrmuirhead !

PR #637, opened 2020-05-25
fix R version to use R old-release

fmichonneau:
@jcoliver this is to fix the R version during the build process to R 3.6.3.


fmichonneau:
specifically requesting the install of git2r seems to address the issue in Travis.

PR #642, opened 2020-06-01
fixed screenshot filenames

maggimars:
This is an updated pull request to one I submitted previously. The previous pr included new screen shots with a more recent version of Rstudio but instead of replacing files I added new ones with the suffix _updated. Now, the files are simply replaced and the filenames are unchanged in the lesson Rmd.


jcoliver:
Thanks for the update, @maggimars !

PR #641, opened 2020-05-31
Adapt "Data structures part 2" to R >= 4.0

csoneson:
This PR adapts the "Data structures part 2" episode to be compliant with R >= 4.0 (in particular, the change in the stringsAsFactors default value). 

This addresses issue #633

PR #645, opened 2020-06-08
Corrections about command output

pablobernabeu:
Corrections: 

1. Updating information about a warning that is not actually produced in the current version of R. Edit: 'In earlier versions of R, the above use of the `rbind()` function ...'

2. Delete notice about an NA that is not actually produced in the current version of R. Notice was: 'Notice that not only did we successfully add a new row, but there is `NA` in the column *coats* where we expected "tortoiseshell" to be.  Why did this happen?'


jcoliver:
Thanks, @pablobernabeu . We are reviewing this now, in light of how we will adapt the lesson to changes introduced in R 4.0.

pablobernabeu:
I just found out that #641 already addressed this.

PR #646, opened 2020-06-10
Separate out for-loop from vectorized solution in the for-loop example

linleyj:
The key point for the vectorization lesson is to "Use vectorized operations instead of loops." When showing how the vectorized operation would look as a for-loop the vectorised operation is included in the same code chunk as the for-loop with no explanation. 

I have put this in a separate code-chunk and added the explanation that this is how the vectorized operation would look like.


jcoliver:
Thanks @linleyj !

PR #651, opened 2020-06-30

pdavid-create:
I made this change because in the April 22-23, 2020 instructor training and breakout sessions we discussed not using dismissive language and the word "basic" may be interpreted as such.  Thus, changing the word in the "Exploring Data Frames" lesson from basic to fundamental is suggested.


jcoliver:
Thanks for the PR, @pdavid-create . On the fence about this one, as "fundamental" carries different weight than does "basic".

jcoliver:
Closing this for now, as we are doing a significant revision of episode 5 due to changes introduced in R 4.0.

PR #654, opened 2020-07-25
Update tidyr functions in the reference

nairps:
updated tidyr functions in the reference


jcoliver:
Excellent! Thanks, @nairps !

PR #662, opened 2020-10-12
Repace iris with mtcars

ErinBecker:
Replaces example using `iris` dataset with example using `mtcars` dataset to avoid using data created by a eugenicist.


naupaka:
Great idea. Thanks for the PR @ErinBecker!

PR #668, opened 2020-11-04
Neat code up with spaces

fursham-h:
I have amended the example code to include spaces in between the selected variables to make it neater and more readable.


jcoliver:
Thanks, @fursham-h !

Issue #603, opened 2020-01-12
Episode 7 (control flow), Challenge 1


mstrimas:
This challenge asks student to print an informative message if there are any records in gapminder for the year 2002. Two solutions are provided, one using `any(gapminder$year == 2002)` (note `any()` isn't introduced until later in that episode) and one much more complicated one involving counting the number of rows for the year 2002. It seems to me the only reasonable way to do this is with `%in%`, which is much simpler and uses a function that was introduced in the previous episode. I propose the solution should be:

```
if (2002 %in% gapminder$year) {
  print("Record(s) for the year 2002 found.")
}
```

I could see an argument for keeping the approach using `any()` as an alternative.

jam2094547:
I think that %in% would be better to introduce as it can be a more powerful tool in the future than any.

jcoliver:
I am inclined to add the `%in%` solution as a third option.

ammaraziz:
I'd like to self-assign this issue if it's acceptable. I also noticed `%in%` is not mentioned in the `reference` page:
https://swcarpentry.github.io/r-novice-gapminder/reference.html

So I might submit this as well during the PR. 

Thanks,

Ammar

PR #672, opened 2020-11-26
Add a third solution to Challenge 1 in Episode 7

agdedrick:
Challenge 1 in Episode 7 asks the student to use an if() command to print a message if any records in gapminder$year are 2002. The lesson currently provides two solutions. In issue #603, @mstrimas suggests a third solution using the %in% operator, which is introduced in an earlier episode, and I added this solution to Challenge 1 material.


jcoliver:
Thanks @agdedrick . We're reviewing this now.

jcoliver:
Thanks again for this PR, @agdedrick . Would you mind updating the PR so the Rmd file is changed (in the _episodes_rmd folder) and the md file (in the _episodes folder) is _not_ changed?

agdedrick:
Whoops, sorry about that! Should be fixed now, @jcoliver.

PR #678, opened 2021-01-06
Add stringsAsFactors = TRUE

jcoliver:
A stopgap measure to ensure `read.csv` reads in character data (as factor) regardless of R version. This PR only adds fixes for episodes 4 and 5; `read.csv` is used elsewhere (episodes 6-14), but should not create significant differences in behavior for different R versions.

### DIFF
--- a/_episodes_rmd/04-data-structures-part1.Rmd
+++ b/_episodes_rmd/04-data-structures-part1.Rmd
@@ -58,7 +58,7 @@ tabby,3.2,1
 We can load this into R via the following:
 
 ```{r}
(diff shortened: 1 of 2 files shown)


naupaka:
Yes this is a great idea. Thanks for making this change. I still have the full re-do on my to-do list.

naupaka (approved):

jcoliver:
Heh. I need to read CONTRIBUTING.md, since I pulled against the wrong branch.

naupaka:
Oops. I should have caught that. Try again and I'll merge?

jcoliver:
Indeed. #684 submitted.

PR #683, opened 2021-01-23
Fix issues from forwarding default values to install.packages explicitly

mlell:
This commit fixed two errors with tooling that arose as I tried to clone
and make the project on my fresh R 4.0.3 installation on Arch Linux.

If install.packages is called with an explicit lib= argument, it expects all
given folders to be readable, which is not the case if the argument is not
provided.

If the repos= argument is not given to install.packages, it can be the special
value @CRAN@, but not if repos= is specified explicitly.
`getOptions(..., default=...)` does not catch this b/c the option is not NULL
but has a special value

### DIFF
--- a/bin/dependencies.R
+++ b/bin/dependencies.R
@@ -1,13 +1,26 @@
 install_required_packages <- function(lib = NULL, repos = getOption("repos", default = c(CRAN = "https://cran.rstudio.com/"))) {
-
-  if (is.null(lib)) {
-    lib <- .libPaths()
+  
+  # The option 'repos' might be set to a special value which confuses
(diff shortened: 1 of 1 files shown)


zkamvar:
Hello,

Sorry for the lateness of this reply. This origin of this particular build file is in https://github.com/carpentries/styles/, and was fixed there in December: https://github.com/carpentries/styles/pull/527

The fix was propagated to this repo recently and should correctly build now.

jcoliver:
Closing this for now, per fix by styles update, but please re-open @mlell if the problem persists.

mlell:
The `repos="@CRAN@"` issue is fixed by #694 and the the `lib=` issue will be fixed with #735.

PR #686, opened 2021-02-03
Update 07-control-flow.Rmd

annajiat:
There is no such function as else().
It is probably a typo.
Changed else() to ifelse().


jcoliver:
Thanks, @annajiat . You make a good point about `else()`, but perhaps the objective would better match the lesson if re-written as 

+ Write conditional statements with `if...else` statements and `ifelse()`.

PR #689, opened 2021-02-12
removed contractions

awesolek2:
made changes recommended in https://github.com/carpentries/carpentries.org/issues/993  for accessibility purposes.


jcoliver:
Thanks, @awesolek2 !

Issue #688, opened 2021-02-12
Episode 8 (ggplot2): reconcile text describing log10 transformation with plot


amzoss:
In [Episode 8](http://swcarpentry.github.io/r-novice-gapminder/08-plot-ggplot2/index.html), the second figure under **Transformations and statistics** is described with the following text:

"The log10 function applied a transformation to the values of the gdpPercap column before rendering them on the plot, so that each multiple of 10 now only corresponds to an increase in 1 on the transformed scale, e.g. a GDP per capita of 1,000 is now 3 on the x axis, a value of 10,000 corresponds to 4 on the x axis and so on. This makes it easier to visualize the spread of data on the x-axis."

It seems to me there are a few inaccuracies. The text suggests that the x axis labels are the log values themselves (3, 4, etc.). Instead, the axis labels are the raw values placed on a log grid (10^3, 10^4, etc.). Perhaps an old version of this plot actually transformed the x values before plotting, instead of applying the scale function, and the text was not updated to match the new plot? That would also explain why the text says "the log10 function" instead of "the scale_x_log10  function".

Proposal: update text to use correct function name and clarify that the labels still match the data, but the coordinate system underneath the plot is now scaled so that the distances between each power of 10 are equal.

Perhaps something like:

"The `scale_x_log10` function applied a transformation to the coordinate system of the plot, so that each multiple of 10 is evenly spaced from left to right. For example, a GDP per capita of 1,000 is the same horizontal distance away from a value of 10,000 as the 10,000 value is from 100,000. This transformation spreads out smaller x values relative to larger values, but the axis still labels the grid according to the original data values."

jcoliver:
Great idea, @amzoss !

jelletreep:
This seems indeed outdated, or at least `log10` in the explanation is not matching the `scale_x_log10` in the example and would be good to fix this. 

I think the suggestion for improvement is great and very clear. Only the last sentence was a bit confusing for me and may be for people not familiar with all concepts, and the latter part of the sentence may be not necessary. My alternative suggestion would be: 

"The `scale_x_log10` function applied a transformation to the coordinate system of the plot, so that each multiple of 10 is evenly spaced from left to right. For example, a GDP per capita of 1,000 is the same horizontal distance away from a value of 10,000 as the 10,000 value is from 100,000. This helps to visualize the spread of the data along the x-axis."

jcoliver:
Thanks for the input, @jelletreep . Care to submit a pull request?

[...]

PR #693, opened 2021-03-04
update reconcile text describing scale log10 transformation with plot

jelletreep:
This pull request follows from issue #688  and revises a figure description in episode 08. The old description mentions a function name that is not used in the example above (log10 versus scale_x_log10. The new description includes the scale_x_log10 function name and tries to clarify what the function does.


jcoliver:
Thanks, @jelletreep !

jcoliver:
Any ideas @fmichonneau or @zkamvar on why this PR results in a build failure? See https://github.com/swcarpentry/r-novice-gapminder/actions/runs/621935318

zkamvar:
Hi @jcoliver, it looks like the styles of the repo needs to be updated since the last update (#671) was done back in November and there have been a number of changes to the validator since then. I will try to initiate a PR.

PR #691, opened 2021-02-26
Fixes gap_wide_betterID and minor style typos

nung001:
as mentioned by @sysilviakim at line 36 of 14-tidyr.Rmd. #687  The first chunk loads and creates the toy example dataset for the lesson and writes it to a new .csv. The code uses the wrong variable, 'gap_wide_betterID' instead of 'gap_wide'. I changed it to 'gap_wide' as  well as applying the styler package which only really "fixed" a two or three indentation "typos".

PR #695, opened 2021-03-09
fixes gap wide typo and adds quotes to title

nung001:
A second attempt at fixing a small typo, changing `gap_wide_betterID` at line 36 to `gap_wide` to write the correct variable to the .csv file. As suggested in the comment on the previous pull request #691  I also added quotes to the tittle and pulled against the master branch this time. I did not use styler to style anything this time as it seemed to introduce some problems. Hope this fixes the issues and thanks for your patience.


jcoliver:
Thanks, @nung001 . We are wrestling with some background infrastructure, and will merge this once that gets figured out.

PR #696, opened 2021-03-10
Update styles

zkamvar:
This should address the issues in https://github.com/swcarpentry/r-novice-gapminder/pull/694#issuecomment-795848265

PR #698, opened 2021-03-15
Update styles 

zkamvar:
This should address the issue from https://github.com/swcarpentry/r-novice-gapminder/pull/696#issuecomment-796265728 and reveal all of the formatting issues that need to be addressed for make check to pass.


zkamvar:
The deed is done!

Issue #721, opened 2021-06-07
Remove dot from facet_grid() function call?


worcjamessmith:
This issue is about the writing data episode (https://swcarpentry.github.io/r-novice-gapminder/11-writing-data/index.html) 

Challenge 1 has this code `p + facet_grid(. ~continent)`. However, `p + facet_grid(~continent)` is sufficient and doesn't require any further explanation beyond what has been used in previous episodes when introducing ggplot (i.e. of the dot). Possibly better to use the simpler option?

jcoliver:
Good point, @worcjamessmith, especially since episode 14 (dplyr) also skips the use of the dot. Want to submit a PR?

jcoliver:
Addressed by #722

PR #722, opened 2021-06-07
Remove unnecessary dot from facet_grid

worcjamessmith:
Dot is not needed and its use is not introduced earlier in the lesson materials. Removed for simplicity.

Addresses issue #721.


jcoliver:
Thanks, @worcjamessmith !

PR #728, opened 2021-07-06
R-novice-gapminder Update 13-dplyr.Rmd (checkout)

Rosiethuypham:
Revised a typo and added more examples to help learners to practice using the functions: select(), filter(), arrange(), and mutate()


jcoliver:
Thanks for these examples, @Rosiethuypham .

PR #735, opened 2021-08-10
2021 08 10 upgrade style

### LINKED ISSUE #733: New version of carpentry/styles should be merged to fix build error due to non-writable directory provided to install.packages()
Hello, 

currently, running `make serve` fails if dependencies must be installed and the R system library is not writable. This is due to line 3 in https://github.com/swcarpentry/r-novice-gapminder/blob/e27ab02afea308c29f15675f38904bab7b90b810/bin/dependencies.R#L1-L5

which forwards the whole contents of .libPaths() to `install.packages(lib=)` if `install_required_packages( lib=NULL,....)`. But `install.packages()` throws an error if not `lib=NULL` or only writable directories are provided. The value of `libPaths()` contains the system library, which might not be writable. This is fixed in [carpentries/styles@da77111](https://github.com/carpentries/styles/commit/da771119b6c4adc61dea3e33786a3c8179600cb1#diff-5f71d0403f53890b9d218b175bf15e5ea7e203d0a1fc92b4f06e977e3912a48a). Merging the newest version of that repository would fix that issue.

fmichonneau:
bring latest version of styles; fixes #733


jcoliver:
Rock on. Thanks, @fmichonneau and @mlell.

PR #736, opened 2021-08-13
add meaningful links

christinemurr:
For web accessibility purposes, rewrote so that the links are meaningful.


jcoliver:
Thanks, @christinemurr !

PR #737, opened 2021-08-17
Lession 4: Remove factors, simpler lists, dedicated names, separate typeof and class, data cleanup challenge. 

mlell:
This is a proposal to restructure some of the lesson about data types. I had the feeling that several things are more complicated than nessecary. I list them below. I have also split up the changes into separate commits that contain some rationale. 

I am unexperienced in the carpentries style of teaching, so if some of these changes contradict some of the best practices, I appreciate if you tell me so that I realize where my own teaching intuition contradicts approaches taken by the carpentries community. 

-----------------------------------

From my subjective perspective, the lesson has the following problems:

* Factors are not necessary for beginners. Most of the time they behave unintuitively and the best is if beginners learn how to avoid them as much as possible. Two packages `vctrs` and `forcats` have been written (in part) to make them less annoying. They are only nessecary for specific areas, like ordering ggplot legends, statistics that are very near to the original mathematics (manually handling design matrices etc.) or handling ordinal information with `<` and `>`, (using `ordered()`. All of these need not be introduced in the base data type lesson. They also do not appear automatically anymore for R > 4.0.1 (stringsAsFactors is FALSE by default now)
* The data type "numeric" is incoherent and outdated. For example `is.numeric` is `TRUE` for integer and double vectors. `double` is more accurate, newer, and corresponds to the output of `typeof()`.
* Many of my colleagues that were unexperienced in R tended to do implicit coercion by accident a lot. In particular they coerced very often back and forth between character and numeric, also with hacks like `if( int_variable == "3")`, after `int_variable` was coerced inadverently to character at some point before. Novices should be suggested some strategy how to spot and avoid unnecessary and implicit data coercing. An exercise that mimics the common task of importing, and cleaning data would be helpful. Also, for reproducibility, the changes made to the data should be documented by a script and not be performed on the original data set.
* The section with the title "Data frames" is incoherent. I do not recognize a clear take home message. Other important info about data.frames is scattered in other sections (for example, the fact that the fundamental structure of a data frame is a list is not found in the data.frame section but in the list section). 
* The concept of names is scattered across different sections, making it appear as if names would be something different for each data type when they are in fact an attribute that is the same for all data structures
* `class()` and `typeof()` are not clearly separated. It is not clear to learners when to use which of the functions. It can also not be inferred, because sometimes they give the same result but sometimes not

The changed lesson can be summarised like this:

**Introduction:** Data.frame columns can be accessed with "$" and can only have one data type (added box on how to remove factors)

**Data types** There are 5 main data types which can be identified using `typeof` and `str`. (remove `class()`)

**Vectors and Coercion**: 
 - **Coercion by combining values**: Combining different data type leads to implicit coercion (largely unchanged)
 - **The type hierarchy**: The types are ordered by what type emerges from a combination of types
 - **Challenge** fill in gaps to import and clean up the broken cat data set (new)

**Some basic vector functions**: How to generate seqences and what does [ do. (remove names)
  - **Challenge**: Generate sequences and perform vectorized ops (remove names)

**Lists**: Lists can organize data of different data types. It has the `[[` operator (big changes)

**Names**: Names are metadata, attached like a label to the object. They allow `["..."]` and `$`  (new/collected from multiple places)
  -  **Accessing elements by name**
  - **Accessing and changing names**
  - **Challenge**: Names are character vectors themselves
  - **Challenge**: Set names on an existing vector (1/2 new, 1/2 moved)
 
**Data frames**: A data frame is a special case of a list. (Heavily changed)
  - **class** the class attribute describes what an object means, the data type describes how an object is constructed.
  - **Challenge**: Different kinds of subsetting (largely unchanged, just removed factors)
  - **Tip: Renaming data frame columns** (moved down here, after names are explained thoroughly)

 **Matrix**: A matrix is a vector plus the "dim" attribute (largely unchanged)

### DIFF
--- a/_episodes_rmd/04-data-structures-part1.Rmd
+++ b/_episodes_rmd/04-data-structures-part1.Rmd
@@ -44,8 +44,8 @@ making a toy dataset in your `data/` directory, called `feline-data.csv`:
 
 ```{r}
 cats <- data.frame(coat = c("calico", "black", "tabby"),
-                   weight = c(2.1, 5.0, 3.2),
-                   likes_string = c(1, 0, 1))
+                    weight = c(2.1, 5.0, 3.2),
+                    likes_string = c(1, 0, 1))
 write.csv(x = cats, file = "data/feline-data.csv", row.names = FALSE)
 ```
 The contents of the new file, `feline-data.csv`:
--- a/_episodes_rmd/04-data-structures-part1.Rmd
+++ b/_episodes_rmd/04-data-structures-part1.Rmd
@@ -693,7 +693,7 @@ pizza_price
 > 3. Test yourself by calling `letter_no["B"]`, which should give you the number
 >    2!
 >
-> > # Solution to Challenge 4
+> > ## Solution to Challenge 4
 > > ```
 > > letter_no <- 1:26   # or seq(1,26)
 > > names(letter_no) <- LETTERS
--- a/_episodes_rmd/04-data-structures-part1.Rmd
+++ b/_episodes_rmd/04-data-structures-part1.Rmd
@@ -14,7 +14,7 @@ objectives:
 keypoints:
 - "Use `read.csv` to read tabular data in R."
 - "The basic data types in R are double, integer, complex, logical, and character."
-- "Use factors to represent categories in R."
+- "Data structures such as data frames or matrices are built on top of lists and vectors, with some added attributes."
 source: Rmd
 ---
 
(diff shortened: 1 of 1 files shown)


jcoliver:
Thanks, @mlell ! This is impressive. It's substantial, so it'll take a little while for us to review it, but there are lots of good ideas in here.

mlell:
OK, @jcoliver, please let me know if there is anything I can do to facilitate your work, like better comparisons before/after or so.... I'm on vacation the next 2 weeks but then I am ready.

naupaka:
@mlell Thanks for this and sorry for the long delay acting on it. Big PRs can be intimidating to review in detail.

I am inclined to merge this; I think on the whole the changes improve and clarify the lesson (and I always felt that this lesson needed some TLC). Your removal of factors is likely controversial, but the callout does mitigate my concern to some extent and I also think it's less of an issue in the R > 4 world where `stringsAsFactors` is now FALSE by default (as you point out).

There is some urgency to acting on this -- we need to merge or close all open PRs before the end of the month, since all Carpentries lessons are switching over to a new branch/CI infrastructure. @matthieu-bruneaux @skanwal Can one or both of you do a read-through to identify if there are any crucial changes that need to be made before merging?

matthieu-bruneaux:
Thanks @mlell for your contribution and @naupaka for looking at this! I reviewed the changes and I think they really help a lot with the flow and clarity of the lesson.

I took the liberty to make a few minor changes to polish some details. LGTM for merging!

matthieu-bruneaux (approved):

mlell:
Thanks for spending the time to look into this, @naupaka and @matthieu-bruneaux , and nice to see that you found it useful!

Issue #742, opened 2021-09-18
Introduce the geom_line() function first, then add the 'by' aesthetic afterwards


K-C-Martin:
Teaching ggplot2 (live coding) for my Instructor demo, I found it a little awkward to demonstrate the difference between geom_point and geom_line while having to simultaneously add in the 'by' aesthetic - I found it smoother to first switch from geom_point to geom_line, then introduce the 'by' aesthetic as a second step.

See https://swcarpentry.github.io/r-novice-gapminder/08-plot-ggplot2/index.html#layers

jcoliver:
Addressed in #743

PR #743, opened 2021-09-18
Introduce geom_line first then include 'by' aesthetic after (ggplot2)

K-C-Martin:
Suggestion for issue #742 ('Introduce the geom_line() function first, then add the 'by' aesthetic afterwards')

The suggested modification plots geom_line without the 'by' aesthetic first, then introduces the 'by' aesthetic to give a better visual appearance to a second plot.

I'm not sure if I've made all the changes necessary: in addition to modifying the ggplot2 Rmd file, I updated one figure in the 'fig' folder and added a new one (the two figures generated by my suggested change).


jcoliver:
Nice addition. Thanks, @K-C-Martin

Issue #719, opened 2021-06-06
Could be interesting to mention the "tidyverse"


InFlamUOB:
Given the chapters on ggplot, dplyr, tidyr... - it could be nice to mention the whole tidyverse universe too - maybe by including just a link in the seeking help chapter e.g "https://www.tidyverse.org" or before teaching ggplot or after tidyr. Knowing that there is a whole universe of as useful packages can be very nice to see and also is a good resource to learn more about those chapters.

naupaka:
I agree -- perhaps this is best done as a callout box just as a reminder for the instructor to mention it.

crschul:
At a workshop I helped with we had a couple of cheatsheets floating around and provided participants with this link to print their own: https://www.rstudio.com/resources/cheatsheets/

jcoliver:
@crschul Indeed. A link to the RStudio page of cheat sheets would be a welcome addition to the ggplot lesson.

mantoniou:
I believe it would also be important to mention that tidyverse packages are **fully compatible with each other**, helping the user to easily combine commands from the different packages (e.g. dplyr, tidyr, ggplot) into a single statement by using the pipe command **%>%**

matthieu-bruneaux:
Thank you for your suggestions @crschul and @mantoniou!

Would anyone like to submit one or two PRs to:

- Add a link to the RStudio page of cheatsheets (for example in the ggplot lesson).
- Update the "Tip: Tidyverse" box in `_episodes_rmd/13-dplyr.Rmd` to mention that the tidyverse packages are designed to work gracefully together and with the pipe?

skanwal:
> Update the "Tip: Tidyverse" box in _episodes_rmd/13-dplyr.Rmd to mention that the tidyverse packages are designed to work gracefully together and with the pipe?

This has been attempted in https://github.com/swcarpentry/r-novice-gapminder/pull/823#pullrequestreview-1376681325

> Add a link to the RStudio page of cheatsheets (for example in the ggplot lesson).
This needs to be still tackled.

naupaka:
Thanks @Phylloxera -- it's in the `episodes` directory in the repo:
https://github.com/swcarpentry/r-novice-gapminder/blob/main/episodes/08-plot-ggplot2.Rmd

Phylloxera:
Added a link to the RStudio page of cheatsheets in the ggplot lesson. #830

PR #748, opened 2021-09-29
Added note on tidyverse

dokato:
Following the approved suggestion from #719 I introduced into dplyr module a mention of Tidyverse family of packages as a tip. I hope this will be useful for the students.


jcoliver:
Thanks for this, @dokato !

Issue #753, opened 2021-10-27
Consider adding Rmd Visual Editor to lesson 15


ekrasch:
The lesson on [knitr using Rmarkdown](https://swcarpentry.github.io/r-novice-gapminder/15-knitr-markdown/index.html) currently does not mention or utilize the visual markdown editor. Consider adding a short note about this tool at the end of the lesson. 

Consider also mentioning the gear icon in the code chunk menu as a way to edit chunk options.

frank0434:
Personally, I have a mixed feeling about using the visual markdown editor. 
I've used it a couple of times since it is realised.
The visual editor does offer this nice rendering for `Rmd` documents. 
But it may take a long time to load if `Rmd` file gets very large. 

I agree with @jcoliver that using this feature might be a stretch. 
However, I reckon it could be part of the [Introduction to Open Data Science with R](https://carpentries.org/blog/2021/03/incubator-spotlight-open-science-r/)?

naupaka:
I in general agree with the mixed feelings about this, but I think a short tip at the end of the lesson might be a good addition.

PR #763, opened 2022-01-09
add tip for visual md editing #753

havablair:
Per discussion in #753  added tip about visual markdown editing in RStudio to lesson 15 "Producing Reports with knitr".  I put it at the end of the lesson as suggested, but I think it could also make sense under the "A bit more markdown" section if we wanted to go that route.


jcoliver:
Thanks, @havablair !

jcoliver:
@havablair Could you re-submit this PR, against the main branch (instead of gh-pages)?

PR #793, opened 2022-08-05

Aariq:
The default for `stringsAsFactors` was changed to `FALSE` in R v4.0.0 so one does not need to remember to set it any longer.

<details>
<summary><strong>Instructions</strong></summary>

Thanks for contributing! :heart:

If this contribution is for instructor training, please email the link to this contribution to
checkout@carpentries.org so we can record your progress. You've completed your contribution
step for instructor checkout by submitting this contribution!

Keep in mind that **lesson maintainers are volunteers** and it may take them some time to
respond to your contribution. Although not all contributions can be incorporated into the lesson
materials, we appreciate your time and effort to improve the curriculum. If you have any questions
about the lesson maintenance process or would like to volunteer your time as a contribution
reviewer, please contact The Carpentries Team at team@carpentries.org.

You may delete these instructions from your comment.

\- The Carpentries
</details>

### DIFF
--- a/_episodes_rmd/05-data-structures-part2.Rmd
+++ b/_episodes_rmd/05-data-structures-part2.Rmd
@@ -130,7 +130,7 @@ str(cats)
 > > ## Solution to Challenge 1
 > > 1. `human_age <- cats$age * 7`
 > > 2. `human_age <- factor(human_age)`. `as.factor(human_age)` works just as well.
-> > 3. `as.numeric(human_age)` yields `1 2 3 4 4` because factors are stored as integers (here, 1:4), each of which is associated with a label (here, 28, 35, 56, and 63). Converting the factor to a numeric vector gives us the underlying integers, not the labels. If we want the original numbers, we need to convert `human_age` to a character vector (using `as.character(human_age)`) and then to a numeric vector (why does this work?). This comes up in real life when we accidentally include a character somewhere in a column of a .csv file supposed to only contain numbers, and forget to set `stringsAsFactors=FALSE` when we read in the data.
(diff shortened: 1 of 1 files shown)


matthieu-bruneaux:
Thanks a lot for noticing this @Aariq! It is always great when we can simplify the lesson material while keeping it up-to-date.

I noticed that we also have several instances in the file `04-data-structures-part1.Rmd` where `stringsAsFactors = TRUE` could be dropped. It would be nice to update this file too so that the PR addresses the `stringsAsFactors` update in one go (and maybe we could have a Tip box after the first place in the lesson where `stringsAsFactors = TRUE` used to be necessary to indicate that older versions of R might cause trouble because of the old default).

I will try to update the PR within the next few days, but feel free to update it before if you would like!

Aariq:
I actually think the appearances of `stringsAsFactors` in `04-data-structures-part1.Rmd` are "fine".  Since the new default for `stringsAsFactors` is `FALSE`, it does need to be set to `TRUE` if you want factors in the data frame.  Personally, I don't think it's a great idea to be teaching non-default usages of `read.csv()` in a first lesson about reading in data, but that change would require a more significant re-writing of the lesson and maybe not everyone agrees.

matthieu-bruneaux:
Sorry for taking so long to get back to you on this @Aariq! And thank you for your explanation, it is true that with the new default the `stringsAsFactors = TRUE` in `04-data-structures-part1.Rmd` are actually required to keep the rest of the lesson working.

I agree with you that teaching to set strings as factors when reading in data might not be the best pedagogical choice. Maybe this is something that we can address in a later update of the lesson.

I took the liberty to update your PR by also removing two unnecessary occurrences of `stringsAsFactors = FALSE` in calls to `data.frame()` in `05-data-structures-part2.Rmd`, since `data.frame()` is not using `stringsAsFactors = TRUE` as default anymore.

Thank you for your contribution @Aariq!

LGTM for merging (cc @skanwal or @naupaka for confirmation).

skanwal (approved):
Thanks for the fix. LGTM as well.

matthieu-bruneaux:
Thank you @skanwal!

PR #802, opened 2022-10-01
Update 10-functions.Rmd - Instructor training checkout contribution

danny-meloncelli:
Functions can be a difficult concept for new programmers. For the temperature conversion functions, I have changed 'temp' to either 'tempF' or 'tempK' so that the student can follow only and know what values are being used at each stage of the function.

This is the first time that I have created a pull request. Hopefully, I have done it right :-)

<details>
<summary><strong>Instructions</strong></summary>

### DIFF
--- a/_episodes_rmd/10-functions.Rmd
+++ b/_episodes_rmd/10-functions.Rmd
@@ -69,8 +69,8 @@ Let's define a function `fahr_to_kelvin()` that converts temperatures from
 Fahrenheit to Kelvin:
 
 ```{r}
-fahr_to_kelvin <- function(temp) {
-  kelvin <- ((temp - 32) * (5 / 9)) + 273.15
+fahr_to_kelvin <- function(tempF) {
+  kelvin <- ((tempF - 32) * (5 / 9)) + 273.15
   return(kelvin)
 }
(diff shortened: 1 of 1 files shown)


naupaka:
@danny-meloncelli thanks for this PR! I agree that changing the parameter name will help clarify the logic. One thing you should fix, though -- we aim to align Carpentries R code style with the [tidyverse style guide](https://style.tidyverse.org/syntax.html#object-names), which advocates for the use of snake_case for all variable names. So if you could update this to use, for example, `temp_f` instead of `tempF`, I think this should be a nice contribution to merge in.

naupaka (changes requested):
@danny-meloncelli thanks for this PR! I agree that changing the parameter name will help clarify the logic. One thing you should fix, though -- we aim to align Carpentries R code style with the [tidyverse style guide](https://style.tidyverse.org/syntax.html#object-names), which advocates for the use of snake_case for all variable names. So if you could update this to use, for example, `temp_f` instead of `tempF`, I think this should be a nice contribution to merge in.

naupaka:
Closing as stale. Feel free to reopen if you have made changes.

PR #807, opened 2022-10-21
added alt text

brynnelliott:
added alt text for https://swcarpentry.github.io/r-novice-gapminder//fig/rmd-09-ch2-sol-2.png

### DIFF
--- a/_episodes_rmd/09-vectorization.Rmd
+++ b/_episodes_rmd/09-vectorization.Rmd
@@ -109,7 +109,7 @@ sum_xy
 > >
 > > Refresh your plotting skills by plotting population in millions against year.
 > >
-> > ```{r ch2-sol}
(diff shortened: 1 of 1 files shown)


ErinBecker:
I think the check failure was due to https://github.com/carpentries/styles/issues/641 and will be fixed with https://github.com/zkamvar/2022-10-20-workflow-bug/blob/main/update-workflows.R which will be deployed today.

zkamvar:
Note: it will not be fixed until https://github.com/swcarpentry/r-novice-gapminder/pull/806 is merged, but even then, because of the way pull requests work, it will still fail after 806 is merged (sorry it's complicated).

zkamvar:
Luckily, your PR is not complex and it will not break anything as long as the maintainer confirms that the syntax is correct.

naupaka:
LGTM

PR #841, opened 2023-07-05
clarify instructions to set working directory

maneesha:
In [this challenge to set the working directory](https://swcarpentry.github.io/r-novice-gapminder/02-project-intro.html#challenge-5) two options are given --  to use the `setwd()` command or to use the option in the `Sessions` menu. 

There are two steps in using the `Sessions` menu.  The second step is displayed as a third item in the list.  This PR puts both steps in the same list item.


matthieu-bruneaux:
Thanks a lot for your contribution @maneesha! My apologies for taking so long to review the PRs from this summer.

PR #850, opened 2023-07-05
include link to loops in shell lesson

maneesha:
As this lesson references loops from the shell lesson, this PR adds a link directly to that lesson.


matthieu-bruneaux:
Thank you @maneesha!

PR #851, opened 2023-07-05
add link to finding things in shell lesson

maneesha:
As this lesson references `grep` and finding things from the shell lesson, this PR adds a link directly to that lesson.


matthieu-bruneaux:
Thank you @maneesha!

PR #855, opened 2023-07-11
run-on sentence

maneesha:


matthieu-bruneaux:
Thanks @maneesha, I took the liberty of making another small edit and then merged.

PR #859, opened 2023-08-14
link to shell loop lessons when mentioned

maneesha:
When the shell loop lessons are mentioned in context of R loops, add link back to the shell loop lessons.


matthieu-bruneaux:
Thank you @maneesha, this was actually solved by merging an earlier PR of yours (#850).

PR #862, opened 2023-08-30
Update slack links

froggleston:
The Carpentries recently updated their Slack domain URL and invite app URL. This PR updates the links to match the new domains.


matthieu-bruneaux:
Thank you for the updated links @froggleston!

PR #866, opened 2023-09-11
Remove reliance on implicit df behavior, #819

### LINKED ISSUE #819: Suggestion: change "Data Frame Manipulation with dplyr" example to not rely on auto-conversion of base R dataframe to atomic vector
The opening example of [Data Frame Manipulation with dplyr](https://swcarpentry.github.io/r-novice-gapminder/13-dplyr/index.html) relies on base R's dataframe behavior to simplify a one column dataframe to an atomic vector: 

```r
gapminder <- read.csv("https://raw.githubusercontent.com/swcarpentry/r-novice-gapminder/gh-

mean(gapminder[gapminder$continent == "Africa", "gdpPercap"])
```

Given that this is one of the implicit behaviors of the datafarme and that this behavior does not happen for tibbles, (below will return a Warning and NA):

```r
library(dplyr)

gapminder <- read.csv("https://raw.githubusercontent.com/swcarpentry/r-novice-gapminder/gh-pages/_episodes_rmd/data/gapminder_data.csv")
gapminder_tidy <- as_tibble(gapminder)

mean(gapminder_tidy[gapminder_tidy$continent == "Africa", "gdpPercap"])
```

I think it's better to make a small change to this example so that it doesn't rely on this implicit behavior of base R dataframe's (particularly given that it won't translate to tidyverse dataframes which are just being introduced). Alternative code for the example could be: 

```r
mean(gapminder_tidy$gdpPercap[gapminder_tidy$continent == "Africa"])
```

p.s. I bumped into this issue when going through this example but changing gapminder data to a tibble so that it would print nicer when going through examples but then bumped into issue with example not working.

brshallo:
Closes #819 

The example I changed relied on implicit behavior of dataframes (which is often better to avoid and doesn't work on tibbles) so I made a slight change to the example that results in the same output but doesn't rely on the implicit dataframe behavior of converting a one column dataframe into a vector.

### DIFF
--- a/episodes/13-dplyr.Rmd
+++ b/episodes/13-dplyr.Rmd
@@ -29,9 +29,9 @@ data by a certain variable(s), or we even calculate summary statistics. We can
 do these operations using the normal base R operations:
 
 ```{r}
-mean(gapminder[gapminder$continent == "Africa", "gdpPercap"])
-mean(gapminder[gapminder$continent == "Americas", "gdpPercap"])
-mean(gapminder[gapminder$continent == "Asia", "gdpPercap"])
+mean(gapminder$gdpPercap[gapminder$continent == "Africa"])
+mean(gapminder$gdpPercap[gapminder$continent == "Americas"])
+mean(gapminder$gdpPercap[gapminder$continent == "Asia"])
 ```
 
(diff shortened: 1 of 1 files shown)


matthieu-bruneaux:
Thank you for your contribution @brshallo! This is very helpful.

PR #868, opened 2023-10-12
Proposed Update to package install at 01-rstudio-intro.Rmd

isaac-jennings:
Dear maintainers, 

I am a SWC Instructor who regularly delivers this R workshop to researchers. Over the course of many workshops in 2023, there have been recurrent issues in relation to lazy loading, and dependency install failure when installing ggplot2 on some machines. 9/10 times, setting the dependencies flag to TRUE explicitly tends to resolve the issues encountered. The other 1/10 times are often new R users setting the library download location to a sync folder i.e. cloud storage (in my experience at least).

For your consideration, and best wishes.

### DIFF
--- a/episodes/05-data-structures-part2.Rmd
+++ b/episodes/05-data-structures-part2.Rmd
@@ -138,14 +138,6 @@ cats <- rbind(cats, cats)
 cats
 ```
 
-But now the row names are unnecessarily complicated. We can remove the rownames,
-and R will automatically re-name them sequentially:
-
-```{r}
-rownames(cats) <- NULL
-cats
-```
-
 :::::::::::::::::::::::::::::::::::::::  challenge
 
 ## Challenge 1
@@ -398,7 +390,7 @@ The object `gapminder` is a data frame with columns
 - Use `cbind()` to add a new column to a data frame.
 - Use `rbind()` to add a new row to a data frame.
 - Remove rows from a data frame.
-- Use `str()`, `summary()`, `nrow()`, `ncol()`, `dim()`, `colnames()`, `rownames()`, `head()`, and `typeof()` to understand the structure of a data frame.
+- Use `str()`, `summary()`, `nrow()`, `ncol()`, `dim()`, `colnames()`, `head()`, and `typeof()` to understand the structure of a data frame.
 - Read in a csv file using `read.csv()`.
 - Understand what `length()` of a data frame represents.
 


matthieu-bruneaux:
Thank you very much for reporting this @isaac-jennings, and also for providing a work-around. It is always very helpful to have feedback from live workshops!

I am a bit reluctant at the moment about adding `dependencies = TRUE` in the learner material for the `"ggplot2"` installation, while not adding it for `"plyr"` nor for `"gapminder"`. I feel this might come as a surprise to the learners, and we would need to add some explanations as to what this argument does and why it is needed in this particular case, but not in the others.

Another aspect is that the default behaviour of `install.packages()` is to fetch and install the dependencies in the `"Depends"`, `"Imports"` and `"LinkingTo"` fields. Using `dependencies = TRUE` tells the function to also install dependencies from the `"Suggests"` field. For ggplot2, the difference in terms of the number of packages to install is quite big on a fresh R install:

- `install.packages("ggplot2")` downloads 23 packages
- `install.packages("ggplot2", dependencies = TRUE)` downloads 110 packages

which also makes me think that having `install.packages("ggplot2", dependencies = TRUE)` in the Learner View for all learners might not be the best solution (assuming that for most learners the ggplot2 install works fine without `dependencies = TRUE`).

An alternative could be to add the `dependencies = TRUE` work-around to the Instructor View only, so that learners who do not experience the issue are not surprised by those instructions but instructors could easily find the work-around to provide it during their workshop if needed. Mentioning the issue with sync folders could also be a very helpful addition to the Instructor View.

If you would like to update your pull request so that the work-around is only visible in the Instructor View, with maybe one or two sentences of explanations for the instructors, I would be happy to review it.

Finally, it would be nice to understand what is the cause of this error during ggplot2 installation, and if it is something which is likely to go away in the future or not. Would it be possible to provide some more details about the error message(s) occurring when installing ggplot2 on the problematic machines (if you remember about them or if you encounter them in future workshops)? I have not encountered this issue yet myself, so I was wondering if it was possible to narrow it down to a specific cause or maybe to a common denominator about the problematic machines (OS? R version? repository used for installation?).

matthieu-bruneaux (changes requested):
As explained in my comment on the PR, I would suggest moving this work-around to the Instructor View only, with a few extra sentences explaining the issue to the instructors and how to solve it if their learners encounter it during their workshop.

PR #885, opened 2024-02-15
Remove plyr episode and update episode numbering

### LINKED ISSUE #540: Remove plyr dependency
From the `plyr` [GitHub repository](https://github.com/hadley/plyr): "`plyr` is retired: this means only changes necessary to keep it on CRAN will be made. We recommend using `dplyr` (for data frames) or `purrr` (for lists) instead." 

I think it would be good practice to remove `plyr` from this lesson. Looking at the files in this repo, I can see that the main place `plyr` is shown is in `12-plyr.Rmd`, with a few references in other places. The tasks in the `plyr` episode are similar to the ones done in `13-dplyr.Rmd`, and I think the tasks in `13-dplyr.Rmd` are actually simpler than the ones in `12-plyr.Rmd`. 

With that in mind, I propose the following changes:
- move `13-dplyr.Rmd` up to be earlier in the sequence 
- do something with `12-plyr.Rmd`. I could imagine a few different possibilities:
    - remove this lesson altogether, to not have more than one episode that feels too similar
    - re-write this lesson to use `dplyr`, updating the tasks to a more modern syntax 
    - replace this lesson with an episode on `purrr`

I am happy to work on a pull request that does any of those things, but I am also very new to the community (I'm submitting this issue as part of the instructor training check-out process!), so if this doesn't sound like a useful contribution, please let me know.

milanmlft:
Fixes #540

Removes the *plyr* episode as it has become obsolete and [*plyr* has been retired](https://github.com/hadley/plyr#status-).
I also updated the version of `igraph` in the package cache, as the older version was giving installation errors when trying to build the lesson locally.


skanwal (approved):
Thanks @milanmlft for tackling this.

I am happy with removing references to plyr.

@naupaka - can you please have a second look to confirm as I notice you've had a look at this issue back in 2019-2020?

naupaka:
I agree it's time to get this removed -- I am still wondering if there is a way to archive it or something. @tobyhodges is that something other lessons have done? If not, then I guess it is reasonable just to remove it and the references to it. I think it's reasonable to separate the removal of plyr and the request for a purrr lesson (which are both part of the discussion on #540).

tobyhodges:
@naupaka we could archive it e.g. by relocating it to the Learners dropdown menu: making it somewhat similar to the extra episode available in [the DC Image Processing curriculum](https://datacarpentry.org/image-processing/). What would be the motivation for keeping it, as opposed to removing it from the lesson altogether? My interpretation of the discussion is that `plyr` has been retired, and the skills and concepts taught in the episode are covered sufficiently elsewhere.

Good point about the request for `purrr` in #540. If/when this is merged, perhaps we can open a new issue to track that aspect? Or keep #540 open but rename it and add a note that `plyr` has been removed?

tobyhodges:
I've opened the pleasingly-numbered #888 to provide a place for further discussion of the possibility to include an episode about `purrr` in the lesson.

@swcarpentry/r-novice-gapminder-maintainers is this ready to merge, or did you want to request something more from @milanmlft?

skanwal:
Happy for this to be merged. Thanks @tobyhodges and @milanmlft

naupaka:
Wow, end of an era. plyr, we knew ye well.

Thanks all!

PR #891, opened 2024-05-16
Ungroup after multi-level grouping to avoid warnings

ndporter:
Add `ungroup()` to all examples grouping by multiple variables to properly return output to ungrouped state. Additionally, add a note in explanation of first multi-variable grouping about need to ungroup.

Previously, output displayed on lesson page in these cases looked like the block below and did not display the summarized data at all (though in RStudio, output should display both).

```
`summarise()` has grouped output by 'continent'. You can override using the
`.groups` argument.
```

*Note: an alternative is (as suggested in warning) to use the `.groups = "drop"` argument in summarise. Both approaches accomplish the same goal. I suggested `ungroup()` because it was previously the only option and has been used in other Carpentries lessons.*

### DIFF
--- a/episodes/12-dplyr.Rmd
+++ b/episodes/12-dplyr.Rmd
@@ -281,12 +281,13 @@ lifeExp_bycountry %>%
 
 :::::::::::::::::::::::::
 
-The function `group_by()` allows us to group by multiple variables. Let's group by `year` and `continent`.
+The function `group_by()` allows us to group by multiple variables. If you don't `ungroup()` after summarizing with multiple grouping variables, you may get a warning telling you the data are still partially grouped. Let's group by `year` and `continent`.
 
 ```{r}
 gdp_bycontinents_byyear <- gapminder %>%
     group_by(continent, year) %>%
-    summarize(mean_gdpPercap = mean(gdpPercap))
+    summarize(mean_gdpPercap = mean(gdpPercap)) %>%
+    ungroup()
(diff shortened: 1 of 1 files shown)


ndporter:
In playing around a little, it may be cleaner to use `.groups="drop"` because  `ungroup()` may still print a warning even if it also renders the output. I'd suggest someone tries out both at least in RStudio with current versions, if not on a local build of the lesson, before deciding which to use.

skanwal (changes requested):
@naupaka - what do you think of this one? Any preferences or favs for grouping data?

naupaka:
Yes I agree this is probably important to address. My instinct is that there needs to be a balance between ungrouping (via either method), which is certainly best practice, and adding too much cognitive overload to new learners. In the old days (before they added in that warning) folks would just leave the data grouped since it only was an issue if the pipe chain was being continued and even then it only led to unexpected behavior in certain instances. In the examples in the lesson, ungrouping is important because the output is stored back into a variable, but I think that's only being done so as to make it display after the image(?). So, the other option here would be to have a callout explaining the warning and leave off the `ungroup()` as well as the variable assignment from the code just to keep the focus on grouping and summarizing, which is the main focus of the content.

I think it could go either way, but I think if we add in either `ungroup()` or `.groups = "drop"` then we probably need slightly more (maybe another sentence or two) of explanation about why this is important.

PR #906, opened 2024-11-08
Change "likes_string" to "likes_catnip"

### LINKED ISSUE #901: Change `likes_string` to something less confusing
### How could the content be improved?

Several times while teaching this lesson, I've had learners stumble over the column in the `cats` data frame called `likes_string`. Especially for learners coming from other programming languages, where there is a data type called "string", this variable name throws them off. They don't immediately think of literal string in relation to cats--instead, they assume that this is a variable called "likes" that has the type "string." Then they get confused by the zeroes and ones, and we have to spend a while explaining that the intention of this part of the lesson is quite different. In fact, when I was taking the lesson for the first time, I also got confused at this part.

I think this is an example of extraneous cognitive load that could be very easily fixed if this variable name were changed to something else with a more obvious meaning, such as `likes_catnip` or `likes_toys` or `likes_playing`. Currently, the variable name is activating learners' prior knowledge in a way that is incorrect and unhelpful.

I'd be happy to submit a PR to fix this, if people like the suggestion! Should be an easy substitution that doesn't change the narrative of the lesson at all.

### Which part of the content does your suggestion apply to?

https://swcarpentry.github.io/r-novice-gapminder/04-data-structures-part1.html

kaijagahm:
Closes #901 

Changed `likes_string` variable in the `cats` data frame to `likes_catnip` instead. This should clear up a frequent point of confusion where learners interpret "string" as referring to a character data type. See further discussion/description in #901.

### DIFF
--- a/episodes/04-data-structures-part1.Rmd
+++ b/episodes/04-data-structures-part1.Rmd
@@ -24,8 +24,8 @@ source: Rmd
 
 ```{r, include=FALSE}
 options(stringsAsFactors = FALSE)
-cats_orig <- data.frame(coat = c("calico", "black", "tabby"), weight = c(2.1, 5, 3.2), likes_string = c(1, 0, 1), stringsAsFactors = FALSE)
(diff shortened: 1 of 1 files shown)


naupaka:
Thanks @kaijagahm!

PR #915, opened 2025-03-31
Fix typo in Data Structures Part 1

blasee:
The `read.csv` function was used in the code block, however it was incorrectly referred to as `read.table`. I have changed the reference from `read.table` to `read.csv` to reflect the correct name of the function used in the code.


skanwal (approved):
Thank you @blasee for catching this.

PR #919, opened 2025-05-01
Change episode number to 14 to Producing Reports With knitr

jsteyn:
The episode number, for Producing Reports With knitr, in the instructor notes is incorrect in two places.


skanwal (approved):
Thank you @jsteyn for fixing the typo.

PR #920, opened 2025-05-03
cats2 should be printed not cats

jsteyn:
This exercise is using the `cats2` object and therefore `cats2` should be printed to view the change and not `cats`


skanwal (approved):
Thanks @jsteyn for fixing this issue.

PR #949, opened 2026-03-16
Fix typo in data frame reference

BRowland-git:
Corrected a typo in one of the challenges that showed data frame as R code instead of the word


cgross95:
Hi @BRowland-git, sorry for the delay in reviewing your merge request! Thanks for the contribution!

In this case, there's some inconsistency throughout that page on the use of "data frame", "data.frame" and "`data.frame`".

I like your suggestion to change "`data.frame`" to "data frame" which is how the concept is introduced. I think it would help to do the same with the remaining instances of "data.frame" on the page. Do you have the capacity to add this change as well?

skanwal (approved):
Thanks - LGTM.

PR #955, opened 2026-06-09
Added callout on `View()` function

david-f1976:
Closes #937 

I concurred with @cgross95 suggestion (https://github.com/swcarpentry/r-novice-gapminder/issues/937#issuecomment-3618725388) in this issue that mention of the `View()` function fits better as a new Tips callout after Challenge 2.

The Tips callout that briefly introduces the data viewer and an example of how to call it from the console. The callout sits *between* Challenge 2 and Challenge 3.

### DIFF
--- a/episodes/05-data-structures-part2.Rmd
+++ b/episodes/05-data-structures-part2.Rmd
@@ -337,11 +337,10 @@ Once opened we can, for example, sort, filter, and search the data frame.
 
 :::::::::::::::::::::::::::::::::::::::::
 
-
-:::::::::::::::::::::::::::::::::::::::  challenge
-
 To make sure our analysis is reproducible, we should put the code
 into a script file so we can come back to it later.
+
+:::::::::::::::::::::::::::::::::::::::  challenge
 
 ## Challenge 3
 
--- a/episodes/05-data-structures-part2.Rmd
+++ b/episodes/05-data-structures-part2.Rmd
@@ -333,7 +333,7 @@ The data viewer can be invoked by using the `View()` function. Typically this wo
 View(gapminder)
 ```
 
-Once opened we can, for example, sort, filter, and search the table.
+Once opened we can, for example, sort, filter, and search the data frame.
 
 :::::::::::::::::::::::::::::::::::::::::
 
--- a/episodes/05-data-structures-part2.Rmd
+++ b/episodes/05-data-structures-part2.Rmd
@@ -317,10 +317,28 @@ gapminder[sample(nrow(gapminder), 5), ]
 
 :::::::::::::::::::::::::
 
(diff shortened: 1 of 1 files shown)


episodes/05-data-structures-part2.Rmd
```diff
@@ -317,11 +317,29 @@ gapminder[sample(nrow(gapminder), 5), ]
 
 :::::::::::::::::::::::::
 
-::::::::::::::::::::::::::::::::::::::::::::::::::
-
 To make sure our analysis is reproducible, we should put the code
 into a script file so we can come back to it later.
```
cgross95, line 320: It looks like there's a slight issue with this part. This is currently set up to be an intro to Challenge 3, but it appears to have moved into an extension of Challenge 2 by moving it above the set of colons.

If you pull this part out and move it to right before Challenge 3, I think it should work well.

episodes/05-data-structures-part2.Rmd
```diff
@@ -317,11 +317,29 @@ gapminder[sample(nrow(gapminder), 5), ]
...
+
+RStudio has a built-in data viewer that allows us to interactively explore a data frame.
+
+The data viewer can be invoked by using the `View()` function. Typically this would be entered in the RStudio console to open the data viewer in the top left pane:
+
+```{r, eval=FALSE, echo=TRUE}
+View(gapminder)
+```
+
+Once opened we can, for example, sort, filter, and search the table.
```
cgross95: I might use "data frame" instead of "table" on this line, just to keep the language consistent.

cgross95 (changes requested):
Thank you for the contribution @david-f1976! We really appreciate you jumping in to help issue #937.

I left a couple of specific comments. One is a minor issue that we'll need resolved to be able to merge, but the other is just a suggestion. Feel free to incorporate or not.

episodes/05-data-structures-part2.Rmd
```diff
@@ -317,13 +317,32 @@ gapminder[sample(nrow(gapminder), 5), ]
...
+Once opened we can, for example, sort, filter, and search the data frame.
+
+:::::::::::::::::::::::::::::::::::::::::
+
 
 :::::::::::::::::::::::::::::::::::::::  challenge
 
+To make sure our analysis is reproducible, we should put the code
+into a script file so we can come back to it later.
+
```
cgross95, line 342: The challenge divs are a little finicky. This intro text should actually be before the

``` md
:::::::::::::::::::::::::::::::::::::::  challenge
```

line above. This will put it between the two challenges, and then the `## Challenge 3` header for the exercise will render correctly.

cgross95 (changes requested):

cgross95 (approved):
Looks great @david-f1976, thanks for moving things around!

